PR #117, opened 2015-11-14
Provide new logging facilities and other helpers

jeromelebleu:
It aims to remove the *display* signals and replace it by the `logging` module. Thanks to some tweaks and configurations, it comes with at least the same features and also more, such as:
 * join logging and messages display in one place to make it easier to control and use
 * select what to log, when and where
 * do not re-invent the wheel, the logging configuration is straight enough to be extended!

Note that it's still compatible with the old-and-soon-deprecated `msignals.display` function in order to facilitate merging and by the way tests.

It also come with some other features such as bash-completion - which still need to be activated per project.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,8 +2,8 @@ Source: moulinette
 Section: python
 Priority: optional
 Maintainer: Jérôme Lebleu <jerome.lebleu@mailoo.org>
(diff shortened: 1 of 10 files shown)


jeromelebleu:
Mmmh.... mmmokay, I take the risk of merging it in a really democratic way.

PR #123, opened 2016-06-23
[enh] update file_not_exist translation to display file path

Psycojoker:
I've tried to respect the usage of quote per languages but I don't understand all of them :/

Required for a PR on yunohost/yunohost


jeromelebleu:
I'll reject your PR, sorry about that. Translations must be updated in Weblate only to avoid conflicts...

PR #124, opened 2016-06-23
Make exception more useful folder

Psycojoker:
Same than https://github.com/YunoHost/moulinette/pull/123 but for folder_not_exists, except we never seems to use this exception according to grep.

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -10,7 +10,7 @@
     "file_exists": "Datei existiert bereits: '{path}'",
     "file_not_exist": "Datei nicht vorhanden: '{path}'",
     "folder_exists": "Ordner existiert bereits: '{path}'",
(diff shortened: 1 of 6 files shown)


jeromelebleu:
I'll reject your PR, sorry about that. Translations must be updated in Weblate only to avoid conflicts...

Psycojoker:
Could you inform me on how to do that?

jeromelebleu:
You have to log in [Weblate](https://translate.yunohost.org), select the right project and component - e.g. YunoHost / moulinette, select the English and French languages and modify the relevant key.

jeromelebleu:
For example, here is what you have to modify for *file_not_exist* in English: https://translate.yunohost.org/translate/yunohost/moulinette/en/?q=file_not_exist&search=substring&context=on&type=all&ignored=False.

PR #125, opened 2016-06-23
[enh] automatically regenerate cache if actionmap.yml is modified

Psycojoker:
For performance reasons I've chosen to use `actionmap.yml` file size as it is super fast to check.

### DIFF
--- a/moulinette/actionsmap.py
+++ b/moulinette/actionsmap.py
@@ -369,10 +369,13 @@ class ActionsMap(object):
             logger.debug("loading actions map namespace '%s'", n)
 
             actionsmap_yml = '%s/actionsmap/%s.yml' % (pkg.datadir, n)
-            actionsmap_yml_stat=os.stat(actionsmap_yml)
-            actionsmap_pkl = '%s/actionsmap/%s-%d-%d.pkl' % (pkg.cachedir, n,
-                                actionsmap_yml_stat.st_size,
-                                actionsmap_yml_stat.st_mtime)
+            actionsmap_yml_stat = os.stat(actionsmap_yml)
+            actionsmap_pkl = '%s/actionsmap/%s-%d-%d.pkl' % (
+                pkg.cachedir,
+                n,
+                actionsmap_yml_stat.st_size,
+                actionsmap_yml_stat.st_mtime
+            )
 
             if use_cache and os.path.exists(actionsmap_pkl):
                 try:
@@ -526,8 +529,10 @@ class ActionsMap(object):
                     os.remove('%s/actionsmap/%s' % (pkg.cachedir, i))
 
             # Cache actions map into pickle file
-            am_file_stat=os.stat(am_file)
-            pkl='%s-%d-%d.pkl' % (n, am_file_stat.st_size, am_file_stat.st_mtime)
+            am_file_stat = os.stat(am_file)
+
+            pkl = '%s-%d-%d.pkl' % (n, am_file_stat.st_size, am_file_stat.st_mtime)
+
             with pkg.open_cachefile(pkl, 'w', subdir='actionsmap') as f:
                 pickle.dump(actionsmaps[n], f)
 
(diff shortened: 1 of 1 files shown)


moulinette/actionsmap.py
```diff
@@ -368,17 +368,23 @@ def __init__(self, parser_class, namespaces=[], use_cache=True,
         for n in namespaces:
             logger.debug("loading actions map namespace '%s'", n)
 
-            if use_cache:
+            actionsmap_yml = '%s/actionsmap/%s.yml' % (pkg.datadir, n)
+            actionsmap_pkl = '%s/actionsmap/%s-%s.pkl' % (pkg.cachedir, n, os.stat(actionsmap_yml).st_size)
```
jeromelebleu: Is relaying on actionmap file size something strength enough for determining its modification? Should we not use `st_mtime` - the time of most recent content modification - instead?

moulinette/actionsmap.py
```diff
@@ -511,8 +517,13 @@ def generate_cache(klass, namespaces=None):
             with open(am_file, 'r') as f:
                 actionsmaps[n] = yaml.load(f)
 
+            # clean old cached files
+            for i in os.listdir('%s/actionsmap/' % pkg.cachedir):
+                if i.endswith(".pkl"):
```
jeromelebleu: I think [`glob.glob`](https://docs.python.org/2/library/glob.html#glob.glob) could be better and more efficient for that, isn't it?

For example:
```python
for path in glob.iglob('%s/actionsmap/*.pkl' % pkg.cachedir):
    os.remove(path)
```

moulinette/actionsmap.py
Psycojoker: It's not cryptographically safe, but it's really safe for our situation (I don't see this being used outside of development) and will lead to less false positives than modification time since it really cached up when the file is effectively modified in 99.9% situations.

If we want we could check both like rsync but I don't see that much additional value added by doing so.

moulinette/actionsmap.py
Psycojoker: If you want to, that really a matter of taste here.

I find the way it's currently written to be easier to understand since it really says what it does (while glob isn't necessarily obvious for everyone) but I don't have a very strong opinion here (while I do prefer code maintainability and readability)

Psycojoker:
Salut,

Suivant nos nouvelles règles d'organisation, je vous re-propose cette PR en considérant, pour moi, que c'est une décision mineur https://github.com/YunoHost/yunohost-project-organization#d%C3%A9cision-mineure

moulinette/actionsmap.py
zamentur: I think situation with only characters change is more than 0,1%. I think use st_mtime in more could be a good idea. For the moment I had not test this code on a ynh instance, but I will and add this small change.

julienmalik: Using mtime in addition seems reasonable at first sight : it makes the cache handling become a lot safer, so we don't need to think about it anymore, or almost.

If the added processing time is not visible, I support using it.

moulinette/actionsmap.py
julienmalik: no clear opinion.
even if I don't know glob prototype and behavior by heart, the code speaks for itself.
it has the advantage of becoming a one-liner.

no idea about performance difference, and no idea whether we really care about performance here (tend to think that we don't care), unless listdir clearly has flaws. but i would guess that glob is more or less listdir+a regex check behind the hood.

Psycojoker: For the principle of testing:

```python
In [10]: %time [with_glob() for x in range(1000000)] and ""
CPU times: user 15.1 s, sys: 3.28 s, total: 18.4 s
Wall time: 18.4 s
Out[10]: ''

In [11]: %time [without_glob() for x in range(1000000)] and ""
CPU times: user 3.55 s, sys: 3.32 s, total: 6.87 s
Wall time: 6.91 s
Out[11]: ''

In [12]: ls
qsd2.pkl  qsd.pkl  ve/
```

Using glob makes it ~ 3 times slower on my computer (no ssd). But yeah, here it really doesn't make any real impact in comparison with everything else we do and I don't think it can be noticeable.

Psycojoker:
I'm ok with the latest changes.

Psycojoker:
Conditions have been reached, initiating merging :robot:

PR #127, opened 2016-11-25
increase cli timeout

Psycojoker:
Hello,

This isn't a great solution but could avoid some bug: it increases the timeout of a cli command from 5 to 30 while we don't have a <code>--wait</code> option.

This is a small decision.

### DIFF
--- a/moulinette/interfaces/cli.py
+++ b/moulinette/interfaces/cli.py
@@ -349,7 +349,7 @@ class Interface(BaseInterface):
                                  lambda a,h: a(password=password))
 
         try:
-            ret = self.actionsmap.process(args, timeout=5)
+            ret = self.actionsmap.process(args, timeout=30)
         except KeyboardInterrupt, EOFError:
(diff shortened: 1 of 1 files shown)


opi:
I don't understand this PR ; could you give us more inputs, use case and benefit of this enlarged timeout ? thx

Psycojoker:
Right now, the slow timeout cause YunoHost commands to randomly fails. This isn't a problem generally but this seems to be happening during installation script and randomly breaking applications installation which is not acceptable at all.

The ideal solution would be to have a general "--wait" option to add to yunohost commands in app scripts (ever better : a global variable for that) but I haven't taken to time to do that yet.

You can find more information on this ticket (there might be other tickets too, it's been a long time reported issue) https://dev.yunohost.org/issues/369

Psycojoker:
Yes, I was really thinking of making another PR to make a possible cli argument and remove it by default, I don't see the point right now (and maybe adding a warning if after 30 seconds there is no answer to tell that it's waiting)

PR #128, opened 2016-11-29
[fix] at installation the cache dir might not be present

Psycojoker:
Hello,

During installation the cache dir might not be present, it is created at this line using a magical helper https://github.com/YunoHost/moulinette/blob/54a50a40b9b678579041e5a8eee2b61e77536aac/moulinette/actionsmap.py#L538

This might that listing a non-existing dir raise an exception and break the installation. This bug is therefor *critical* and needs to be merged (or another solution) before we make a release.

This is a minor decision.

### DIFF
--- a/moulinette/actionsmap.py
+++ b/moulinette/actionsmap.py
@@ -523,10 +523,12 @@ class ActionsMap(object):
             with open(am_file, 'r') as f:
                 actionsmaps[n] = yaml.load(f)
 
-            # clean old cached files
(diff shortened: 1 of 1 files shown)


Psycojoker:
I can chose to bypass small obvious fix in the future if you want but I'm still interested in some review macanism.

opi:
imho, we can merge it asap, as it is a critical bug fix. (but I don't want to bypass any decision mechanism)

Psycojoker:
I'm going to merge it then, and consider the decision done once the condition is reach, otherwise we'll revert.

PR #129, opened 2016-12-02
Wait by default on cli

Psycojoker:
Hello,

Due to numerous random bugs (a lot of random installation errors) and no benefit added, I propose that we remove the timeout mecanism from the CLI interface. This PR do this.

### DIFF
--- a/moulinette/core.py
+++ b/moulinette/core.py
@@ -534,7 +534,7 @@ class MoulinetteLock(object):
             lock
 
     """
-    def __init__(self, namespace, timeout=0, interval=.5):
+    def __init__(self, namespace, timeout=None, interval=.5):
         self.namespace = namespace
         self.timeout = timeout
         self.interval = interval
@@ -573,7 +573,7 @@ class MoulinetteLock(object):
                     self._lock()
                     break
 
-            if (time.time() - start_time) > self.timeout:
+            if self.timeout is not None and (time.time() - start_time) > self.timeout:
                 raise MoulinetteError(errno.EBUSY,
                                       m18n.g('instance_already_running'))
             # Wait before checking again
--- a/moulinette/interfaces/cli.py
+++ b/moulinette/interfaces/cli.py
@@ -322,7 +322,7 @@ class Interface(BaseInterface):
 
         self.actionsmap = actionsmap
 
-    def run(self, args, output_as=None, password=None):
+    def run(self, args, output_as=None, password=None, timeout=None):
         """Run the moulinette
 
         Process the action corresponding to the given arguments 'args'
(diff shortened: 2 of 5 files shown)


Psycojoker:
We also need to merge to PR https://github.com/YunoHost/yunohost/pull/200 for this one to take action.

Psycojoker:
Oh btw, I'm using os.environ which is super dirty, but that's the current way to do so (it's already used for bypass_lock) because there is currently no other way to do so in the current architecture.

moulinette/core.py
```diff
@@ -554,6 +554,11 @@ def acquire(self):
         """
         start_time = time.time()
 
+        if os.environ.get("YUNOHOST_CLI_WAIT") == "yes":
```
julienmalik: moulinette is supposed to be more or less independent of yunohost.

Psycojoker:
Ah no, the break was due to one of my test. I think the admin interface timeout is supposed to be 30 (and not 0), I've changed it.

moulinette/interfaces/cli.py
```diff
@@ -335,6 +335,7 @@ def run(self, args, output_as=None, password=None):
                 - plain: return a script-readable output
                 - none: do not output the result
             - password -- The password to use in case of authentication
+            - timeout -- Timeout this command after 30 seconds if the lock isn't acquired
```
julienmalik: also, reading the code, 0 does not mean "no timeout" as one could expect, so might be worth mentioning here.

moulinette/core.py
```diff
@@ -573,7 +573,7 @@ def acquire(self):
                     self._lock()
                     break
 
-            if (time.time() - start_time) > self.timeout:
+            if self.timeout is not None and (time.time() - start_time) > self.timeout:
```
julienmalik: maybe add a test for self.timeout==0 so that 0 means "no timeout" ? (see below)

moulinette/interfaces/cli.py
Psycojoker: Ah no, it's the value passed to the argument, forgot to update the doc

It would be

    yunohost domain list --timeout 42

moulinette/core.py
julienmalik: ok for `None`, but what happens when using `--timeout 0` ?

julienmalik: I was just wondering if it makes sense that timeout 0 is **also** a case for "no timeout"...

Psycojoker: For me, it makes more sens to mean "don't wait at all" but this point could be discussed as I have no strong opinion on it.

julienmalik: Now that I understood the meaning of life, please disregard my remarks about the timeout = 0 case

Psycojoker:
Updated following comments on the doc.

julienmalik:
Thanks LGTM

PR #131, opened 2016-12-07
Unfinished process bug

Psycojoker:
Hello,

This is a fix for this super long ticket https://dev.yunohost.org/issues/654. After (long) investigations by @maniackcrudelis, we end up finding that add a `time.sleep(1)` solves the issue on his testing environment.

This PR had a timeout mecanism to make this approach a bit more robust.

This bug is critical for our CI.

**I haven't tested this PR yet, I'm too tired for that and need to sleep for work.**

### DIFF
--- a/moulinette/utils/process.py
+++ b/moulinette/utils/process.py
@@ -81,6 +81,17 @@ def call_async_output(args, callback, **kwargs):
             time.sleep(.1)
     stdout_reader.join()
     stdout_consum.join()
+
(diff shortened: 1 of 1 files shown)


Psycojoker:
Hello,

This code has been tested by @maniackcrudelis during several hours on the setup that was triggering this bug. I consider this patch read for merge and would be interested in reviews (but since that much time have passed I'm tempted to merge it directly).

Psycojoker:
<3

PR #133, opened 2016-12-22
log.SUCCESS Not Supported

austin1029:
AttributeError: 'module' object has no attribute 'SUCCESS'


Psycojoker:
Indeed, that's fixed, thx @austin1029 for your contribution :)

PR #138, opened 2017-03-27
[fix] Show positional arguments first in --help / usage

alexAubin:
### Status

Ready for reviews / feedback

### Problem

This is a proposal to fix what's described in https://dev.yunohost.org/issues/867. The --help / usage shown for some commands is confusing and sometimes does not work as expected

```bash
$ yunohost app addaccess --help
usage: yunohost app addaccess [-h] [-u [USERS [USERS ...]]] apps [apps ...]
```
```bash
$ yunohost app addaccess -u some_user some_app
yunohost app addaccess: error: too few arguments
```
```bash
$ yunohost app addaccess some_app -u some_user
[works]
```

This is a bug / inconsistent behavior in `argparse`, discussed [here](http://stackoverflow.com/questions/5854920/argparse-incorrect-order-of-positional-and-optional-parameters), [here](http://bugs.python.org/issue9338) (2010 :sob:) and in other places

Digging into this, I don't understand the reason why positional arguments are shown AFTER optional arguments. This makes some "usage" messages confusing / not intuitive, such as 

```bash
$ yunohost user update --help
usage: yunohost user update [*~list of many many options ...~*] username
```

### Solution

Argparse allows to specify a [custom formatter](http://www.enseignement.polytechnique.fr/informatique/INF478/docs/Python3/library/argparse.html#formatter-class) to build the help/usage message.

I copy-pasted the original code from the argpaste.HelpFormatter class and tweaked it to display positional argument first (inspired from [this thread](http://stackoverflow.com/questions/26985650/argparse-do-not-catch-positional-arguments-with-nargs/26986546#26986546))

### Testing

Compare outputs (with/without applying the PR) or some help messages such as : 

```
yunohost app addaccess --help
yunohost user create --help
yunohost user update --help
```

### DIFF
--- a/moulinette/interfaces/__init__.py
+++ b/moulinette/interfaces/__init__.py
@@ -556,7 +556,7 @@ class ExtendedArgumentParser(argparse.ArgumentParser):
 
 
 # This is copy-pasta from the original argparse.HelpFormatter :
-# https://svn.python.org/projects/python/trunk/Lib/argparse.py
+# https://github.com/python/cpython/blob/1e73dbbc29c96d0739ffef92db36f63aa1aa30da/Lib/argparse.py#L293-L383
 # tweaked to display positional arguments first in usage/--help
 #
 # This is motivated by the "bug" / inconsistent behavior described here :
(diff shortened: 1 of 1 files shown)


alexAubin:
As suggested by Bram, I added some "TWEAK" flags to show where I changed stuff with respect to original code.

alexAubin:
Planning to merge this around April 2nd if there's no opposition

PR #141, opened 2017-05-03
[enh] More helpers for common IO operations

alexAubin:
This is a continuation of https://github.com/YunoHost/yunohost/pull/274 as I realized that what I wanted to implement was already partly in the moulinette and it makes more sense to have them here as they are independent from Yunohost itself.

Here's the initial Problem/Solution description :

>### Problem
>
>Follow-up of a discussion we had two or three times already : we keep needing to read / write json files, change file permissions or downloading stuff at various places in the code. This lead to either duplicated code, or different pieces of code that don't handle exceptions consistently. The current situation also discourage to handle exceptions correctly because there are simply too many of them to check (e.g. when doing a `requests.get()`)
>
>### Proposed solution
>
>Introduce <s>io.py</s>*helpers* that contains low-level functions for all common operations such as : 
>- read/write files
>- read/write json
>- download text/json
>- change file permissions *(Note : already in Moulinette in fact)*
>- execute ~raw bash commands *(Note : already in Moulinette in fact)*
>- ???

I implemented unit tests (though they don't cover all the previously existing Moulinette utils for now). I also tweaked the behavior of the existing `check_commands` (which I renamed into `call_commands` to be more meaningful). It's only used in `yunohost.firewall` one or two times so shouldn't be so troublesome to propagate this.

### DIFF
--- a/moulinette/utils/filesystem.py
+++ b/moulinette/utils/filesystem.py
@@ -1,12 +1,133 @@
 import os
 import errno
 import shutil
+import json
+import grp
 from pwd import getpwnam
 
 from moulinette.core import MoulinetteError
 
-
 # Files & directories --------------------------------------------------
+
+def read_file(file_path):
+    """
+    Read a regular text file
+
+    Keyword argument:
+        file_path -- Path to the json file
+    """
+    assert isinstance(file_path, str)
+
+    # Check file exists
+    if not os.path.isfile(file_path):
+        raise MoulinetteError(errno.ENOENT,
+                              m18n.g('file_not_exist', path=file_path))
+
+    # Open file and read content
+    try:
+        with open(file_path, "r") as f:
+            file_content = f.read()
+    except IOError as e:
+        raise MoulinetteError(errno.EACCES,
+                              m18n.g('cannot_open_file', file=file_path))
+    except Exception as e:
+        raise MoulinetteError(errno.EIO,
+                              m18n.g('error_reading_file',
+                                     file=file_path, error=str(e)))
+
+    return file_content
+
+
+def read_json(file_path):
+    """
+    Read a json file
+
+    Keyword argument:
+        file_path -- Path to the json file
+    """
+
+    # Read file
+    file_content = read_file(file_path)
+
+    # Try to load json to check if it's syntaxically correct
+    try:
+        loaded_json = json.loads(file_content)
+    except ValueError:
+        raise MoulinetteError(errno.EINVAL,
+                              m18n.g('corrupted_json', ressource=file_path))
+
+    return loaded_json
+
+
+def write_to_file(file_path, data, file_mode="w"):
+    """
+    Write a single string or a list of string to a text file.
+    The text file will be overwritten by default.
+
+    Keyword argument:
+        file_path -- Path to the output file
+        data -- The data to write (must be a string or list of string)
+        file_mode -- Mode used when writing the file. Option meant to be used
+        by append_to_file to avoid duplicating the code of this function.
+    """
+    assert isinstance(data, str) or isinstance(data, list)
+    assert not os.path.isdir(file_path)
+
+    # If data is a list, check elements are strings and build a single string
+    if not isinstance(data, str):
+        for element in data:
+            assert isinstance(element, str)
+        data = '\n'.join(data)
+
+    try:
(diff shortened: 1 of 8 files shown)


moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+        raise MoulinetteError(errno.ENOENT,
+                              m18n.g('file_not_exist', path=file_path))
+
+    # Open file and read content
+    try:
+        with open(file_path, "r") as f:
+            file_content = f.read()
+    except IOError as e:
+        raise MoulinetteError(errno.EACCES,
+                              m18n.g('cannot_open_file', file=file_path))
```
Psycojoker: we should also supply the error to know why the file can't be read

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+    except Exception as e:
+        raise MoulinetteError(errno.EIO,
+                              m18n.g('error_reading_file',
```
Psycojoker: same

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+    """
+
+    # Read file
+    file_content = read_file(file_path)
+
+    # Try to load json to check if it's syntaxically correct
+    try:
+        loaded_json = json.loads(file_content)
+    except ValueError:
+        raise MoulinetteError(errno.EINVAL,
```
Psycojoker: same, otherwise we can't guess where it's corrupted :x

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+    Write a single string or a list of string to a text file.
+    The text file will be overwritten by default.
+
+    Keyword argument:
+        file_path -- Path to the output file
+        data -- The data to write (must be a string or list of string)
+        file_mode -- Mode used when writing the file. Option meant to be used
+        by append_to_file to avoid duplicating the code of this function.
+    """
+    assert isinstance(data, str) or isinstance(data, list)
```
Psycojoker: will fail is data is unicode, should test with "base_string"

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+        for element in data:
+            assert isinstance(element, str)
+        data = '\n'.join(data)
+
+    try:
+        with open(file_path, file_mode) as f:
+            f.write(data)
+    except IOError as e:
+        raise MoulinetteError(errno.EACCES,
+                              m18n.g('cannot_write_file', file=file_path))
```
Psycojoker: same for exception

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+    except Exception as e:
+        raise MoulinetteError(errno.EIO,
+                              m18n.g('error_writing_file',
```
Psycojoker: same

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+
+    Keyword argument:
+        file_path -- Path to the output json file
+        data -- The data to write (must be a dict or a list)
+    """
+
+    # Assumptions
+    assert isinstance(file_path, str)
+    assert isinstance(data, dict) or isinstance(data, list)
+    assert not os.path.isdir(file_path)
```
Psycojoker: if you want to be super rigorous you can test if the directory exists.

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+
+    # Write dict to file
+    try:
+        with open(file_path, "w") as f:
+            json.dump(data, f)
+    except IOError as e:
+        raise MoulinetteError(errno.EACCES,
+                              m18n.g('cannot_write_file', file=file_path))
```
Psycojoker: same

moulinette/utils/filesystem.py
```diff
@@ -1,13 +1,134 @@
...
+    except Exception as e:
+        raise MoulinetteError(errno.EIO,
+                              m18n.g('_error_writing_file',
```
Psycojoker: same

PR #142, opened 2017-07-03
[fix] Correctly handle error 500

alexAubin:
### Problem

At the moment, the moulinette does not properly handle exception when called with the API. This trigger an obscure "500 Internal Server Error" on the webadmin.

### Solution

Add an exception handling and feed the web admin with the details/traceback of what happened.

Related PR on Moulinette : https://github.com/YunoHost/yunohost-admin/pull/166

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -483,7 +483,6 @@ class HTTPErrorResponse(HTTPResponse):
 
 
 def error_to_response(error):
-
     """Convert a MoulinetteError to relevant HTTP response."""
     if error.errno == errno.EPERM:
         return HTTPForbiddenResponse(error.strerror)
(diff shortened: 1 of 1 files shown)


moulinette/interfaces/api.py
```diff
@@ -387,6 +387,15 @@ def process(self, _route, arguments={}):
             ret = self.actionsmap.process(arguments, timeout=30, route=_route)
         except MoulinetteError as e:
             raise error_to_response(e)
+        except Exception as e:
+            if isinstance(e, HTTPResponse):
+                raise e
+            import traceback
```
Psycojoker: I wonder if we shouldn't do this import at the beginning of the file.

alexAubin: Yea, I dunno what's the best practice about that... On one hand it's cleaner to import it at the beginning, on the other hand it's a module only used in edge cases so it's useless to load it in 99.9% of cases... Personnaly I don't have a strong opinion about this, I just implemted this that way because we have this in other part of the code...

Psycojoker: That's pretty fast to load since it's a small stl module. But let's not care about that, that's fine to have it here.

Psycojoker:
Not tested, code looks good.

alexAubin:
Flagging this as ready to merge / merging in a few days. Please complain if you intend to review or want to request changes.

PR #144, opened 2017-07-24
try to make ActionsMap._construct_parser code more readable

Psycojoker:
I recommend to read this commits by commits as a lot of things have been modified.

I've tried to make "one action per commit" to make the workflow more understandable.

If that's too much at once we can try to split into several batch of commits.

Normally I haven't break anything (yunohost-{cli,api} works for me with this refactoring in ynh-dev).

Still, it's not perfect, the code makes a lot way too much interconnections and is full of magic and "out of nowhere" stuff so we'll have to move step by step and this is a first one.

I'm still impressed by all the things that moulinette does, this isn't a trivial problem.

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -548,7 +548,7 @@ class ActionsMapParser(BaseActionsMapParser):
             return [name.replace('--', '@', 1)]
         return [name.replace('-', '@', 1)]
 
-    def add_global_parser(self, **kwargs):
+    def get_global_parser(self, **kwargs):
         raise AttributeError("global arguments are not managed")
 
     def add_category_parser(self, name, **kwargs):
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -80,6 +80,24 @@ class _HTTPArgumentParser(object):
     def get_default(self, dest):
         return self._parser.get_default(dest)
 
+    def add_arguments(self, arguments, extraparser, format_arg_names=None, validate_extra=True):
+        for argument_name, argument_options in arguments.items():
+            # will adapt arguments name for cli or api context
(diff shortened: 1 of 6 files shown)


Psycojoker:
I haven't solved this travis warning things as it is coming from a previous PR (yes, it's you @alexAubin :yum:) and that would make things confusing here.

alexAubin:
>I haven't solved this travis warning things as it is coming from a previous PR (yes, it's you @alexAubin :yum:) and that would make things confusing here.

Sorry 'bout dat  :anguished:, should be fixed by 51ea0698018555f75703144bbc4b661752c58545  :grin:

moulinette/interfaces/api.py
```diff
@@ -80,6 +80,24 @@ def set_defaults(self, **kwargs):
     def get_default(self, dest):
         return self._parser.get_default(dest)
 
+    def add_arguments(self, arguments, extraparser, format_arg_names=None, validate_extra=True):
```
alexAubin: Uh, is this needed in api.py ? o.O"

alexAubin: Ah, yes it is, because the _HTTPArgumentParser is what's returned by add_action_parser. Hm... Wondering if that's worth factorizing this in a mother class like 'MoulinetteActionParser' to avoid copy-pasta of the add_arguments... But maybe that's good enough.

Psycojoker: Yes, very possibly, but right now I was having a very "make this only function readable" approach, a lot of things should be simplified in moulinette but I didn't want to put too much stuff in one PR.

alexAubin: Alrighty :+1:

alexAubin:
Flagging this as ready to merge / merging in a few days. Please complain if you intend to review or want to request changes.

PR #152, opened 2017-08-15
[fix] auto upgrade admin password to sha-512 on login

Psycojoker:
Related to hash PR on yunohost.

Related to https://github.com/YunoHost/yunohost/pull/350


alexAubin:
Not tried yet, but ideally, I wish we could have this as a manual migration in Yunohost (manual because we need to ask the admin to enter the password). This would avoid to have this piece of code which contains duplicate from YunoHost and will only be ran once on existing installs...

Psycojoker:
> Edit : this could be as simple as asking the password then calling tools_adminpw(). I think we just need the 'manual migration' thing.

That would be great but it's not ready yet and not available in the admin :(

PR #154, opened 2017-09-06
Allow to give lock to multiple processes

alexAubin:
### Problem

We want to be able to manage several locks. Also a command waiting to acquire the lock should keep checking if one of its parent got it in the meantime.

This is motivated by https://github.com/YunoHost/yunohost/pull/367

### Solution

The lock file can now contain several lock (on separate lines) and keep checking in `acquire()` if one of its parent got the lock (instead of just once in `__enter__()`).

### Validation

- [x] Principle agreement 1/2 : ljf
- [x] Quick review 1/1 : ljf
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1? :

### DIFF
--- a/moulinette/core.py
+++ b/moulinette/core.py
@@ -446,16 +446,17 @@ class MoulinetteLock(object):
             if self._is_son_of_locked():
                 return
 
-            lock_pid = self._lock_PID()
+            lock_pids = self._lock_PIDs()
 
-            if lock_pid is None:
+            if lock_pids == []:
                 self._lock()
                 break
             elif not self._stale_checked:
                 self._stale_checked = True
-                # Delete stale lock file
-                if not lock_pid or not os.path.exists(
-                        os.path.join('/proc', lock_pid, 'exe')):
+                # Check locked process still exist and take lock if it doesnt
+                # FIXME : what do in the context of multiple locks :|
+                first_lock = lock_pids[0]
+                if not os.path.exists(os.path.join('/proc', str(first_lock), 'exe')):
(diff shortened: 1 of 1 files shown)


moulinette/core.py
```diff
@@ -442,16 +442,21 @@ def acquire(self):
...
+            if lock_pids == []:
                 self._lock()
                 break
             elif not self._stale_checked:
                 self._stale_checked = True
-                # Delete stale lock file
-                if not lock_pid or not os.path.exists(
-                        os.path.join('/proc', lock_pid, 'exe')):
+                # Check locked process still exist and take lock if it doesnt
+                # FIXME : what do in the context of multiple locks :|
```
zamentur: We might check `not os.path.exists(os.path.join('/proc', str(PID), 'exe'))` for each PID in the `_lock_PIDs` list ?

moulinette/core.py
```diff
@@ -442,16 +442,21 @@ def acquire(self):
         start_time = time.time()
 
         while True:
-            lock_pid = self._lock_PID()
 
-            if lock_pid is None:
+            lock_pids = self._lock_PIDs()
+
+            if self._is_son_of(lock_pids):
+                return
```
zamentur: Why use `return` here and `break` below ?

alexAubin: Because if we are a son of the locked process, we don't want to actually acquire the lock.

Whereas (in the next lines), if there are no current locks, we want to take the lock (with `self._lock()`)

PR #155, opened 2017-11-04
[enh] Optionnal stdinfo communication channel when running commands

alexAubin:
### Problem

See https://github.com/YunoHost/yunohost/pull/383

### Solution

This took much longer and is much less simple/elegant than my original POC -.-

- Have an optionnal stdinfo argument for `call_async_output` (used by hook_exec). It will create a FIFO/'named' pipe
- The corresponding FIFO file has to be handled with `os.open()` and not just `open()` (need non-blocking opening...)
- ... and we need to read stuff with a low-level `os.read()` :/ in AsynchronousFileReader (the stuff that also reads stdout, stderr asynchronously and feed it to logger.warning() and so on..)
- (Minor : fix the display of 'Info' prefix in --verbose, key was missing)

### How to test

See https://github.com/YunoHost/yunohost/pull/383


zamentur:
tested , it works with a ynh-dev new version : https://github.com/YunoHost/ynh-dev/pull/41

PR #158, opened 2018-04-21
[enh] Lazy-load some modules to improve performances

pitchum:
While troubleshooting performance issues on yunohost I found out that some package imports were very slow and I think loading time is valuable information for debugging.

This PR makes this information visible when using `--debug` command line switch.
I also moved a slow import of module *requests*. Although I'm not sure this one currently improves perfs, it may be the case in the future.


alexAubin:
Proposal to merge this in 3 days if there's no opposition

PR #159, opened 2018-04-24
[mod] remove catch-all try and provide real exception

Psycojoker:
Avoid catch-all, they are a nightmare to debug :(


Psycojoker:
Labelling that as a microdecision for now.

PR #165, opened 2018-06-23
[mod] uses thread to allow callback to modify the context

Psycojoker:
This is needed for config-panel to be able to work.

### DIFF
--- a/moulinette/utils/stream.py
+++ b/moulinette/utils/stream.py
@@ -1,3 +1,5 @@
+import threading
+
 from multiprocessing.process import Process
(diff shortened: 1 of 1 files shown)


Psycojoker:
Well, it was working like a charm for the CLI interface but break the message streaming for the admin for a totally unknown reason, I need to fix that before merging it >_>

But I really need a way for async callbacks to be able to communicate with the main thread and process prevent that :(

Psycojoker:
I found a better way to do that.

Also this PR was crap, the real things to do was in my stash queue for some reason.

PR #168, opened 2018-08-06
Dont crash on failed to format translation string

Psycojoker:
Fix the describe situation here https://github.com/YunoHost/yunohost/pull/512#discussion_r207753553

Tested in yunohost-tools-shell.

```
In [1]: m18n.n('app_location_unavailable')
Out[1]: 'This url is not available or conflicts with the already installed app(s):\n{apps:s}'
```

For some reason warning won't appear in yunohost shell :C

### DIFF
--- a/moulinette/core.py
+++ b/moulinette/core.py
@@ -89,13 +89,20 @@ class Translator(object):
 
         """
         if key in self._translations.get(self.locale, {}):
-            return self._translations[self.locale][key].encode('utf-8').format(*args, **kwargs)
+            try:
+                return self._translations[self.locale][key].encode('utf-8').format(*args, **kwargs)
+            except KeyError as e:
+                logger.warning("Failed to format translated string '%s' with error: %s" % (key, e))
 
         if self.default_locale != self.locale and key in self._translations.get(self.default_locale, {}):
             logger.info("untranslated key '%s' for locale '%s'",
                         key, self.locale)
 
-            return self._translations[self.default_locale][key].encode('utf-8').format(*args, **kwargs)
+            try:
+                return self._translations[self.default_locale][key].encode('utf-8').format(*args, **kwargs)
+            except KeyError as e:
+                logger.warning("Failed to format translatable string '%s' with error: %s" % (key, e))
+                return key
 
         logger.exception("unable to retrieve key '%s' for default locale '%s'",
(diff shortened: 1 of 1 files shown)


alexAubin:
>For some reason warning won't appear in yunohost shell :C

I came to understand that logger.exception() only show up with --debug ? (At least in the yunohost shell)

Otherwise, that works as expected :+1:

Psycojoker:
> I came to understand that logger.exception() only show up with --debug ? (At least in the yunohost shell)

That would be a bit problem as exception is supposed to be way superior to info/debug and even warning in the logger hierarchy :o 

I tried it before with warning but couldn't get them to show up neither :/

moulinette/core.py
```diff
@@ -88,14 +88,23 @@ def translate(self, key, *args, **kwargs):
...
-            return self._translations[self.locale][key].encode('utf-8').format(*args, **kwargs)
+            try:
+                return self._translations[self.locale][key].encode('utf-8').format(*args, **kwargs)
+            except KeyError as e:
+                logger.exception("Failed to format translated string '%s' with error: %s" % (key, e))
+                failed_to_format = True
 
-        if self.default_locale != self.locale and key in self._translations.get(self.default_locale, {}):
+        if failed_to_format or (self.default_locale != self.locale and key in self._translations.get(self.default_locale, {})):
             logger.info("untranslated key '%s' for locale '%s'",
```
alexAubin: Hm that's no big deal, but this message is sometimes displayed even if the key is actually translated ...

for instance with locale as Fr : 

```
>>> m18n.n("user_unknown", app="test")
15044 WARNING Failed to format translated string 'user_unknown' with error: 'user'
15044 INFO untranslated key 'user_unknown' for locale 'fr'
15044 WARNING Failed to format translatable string 'user_unknown' with error: 'user'
'Utilisateur \xc2\xab {user:s} \xc2\xbb inconnu'
```

and a bit weirded with locale as En since it's the default locale : 

```
>>> m18n.n("user_unknown", app="test")
8407 WARNING Failed to format translated string 'user_unknown' with error: 'user'
8408 INFO untranslated key 'user_unknown' for locale 'en'
8408 WARNING Failed to format translatable string 'user_unknown' with error: 'user'
'Unknown user: {user:s}'
```

PR #169, opened 2018-08-31
Convert bool-like string arguments to actual bool

alexAubin:
### The problem

I need to be able to send a boolean argument '`force':True` (equivalent to `--force`) from the webadmin in the context of a POST request.

However, turns out that transmitting booleans in POST data is [not 100% trivial since only strings are transmitted anyway](https://stackoverflow.com/questions/3654454/boolean-variables-posted-through-ajax-being-treated-as-strings-in-server-side). So far we did not need such a thing...

### Proposed solution

I could do some dirty hack in my command API but that wouldnt be pretty for CLI users ... So instead I propose this fix were, if an arg looks like a boolean (i.e. `"true"` or `"false"`) it is converted to `True/False`

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -402,6 +402,14 @@ class _ActionsMapPlugin(object):
             - arguments -- A dict of arguments for the route
 
         """
+
+        # Convert POST arguments to boolean if they look like one...
+        for k, v in arguments.items():
+            if v.lower() == "true":
+                arguments[k] = True
+            elif v.lower() == "false":
+                arguments[k] = False
+
         try:
             ret = self.actionsmap.process(arguments, timeout=30, route=_route)
         except MoulinetteError as e:


zamentur:
If someone try to put false or true as a password, i think it will display a warning and put password as password ...
https://github.com/YunoHost/moulinette/blob/stretch-unstable/moulinette/actionsmap.py#L114

zamentur:
@alexAubin  What about this one ?

alexAubin:
Meh, i dunno, https://github.com/YunoHost/yunohost-admin/pull/209 is in standby while we progress on other things so we'll see... But to me the edge case of "someone putting 'true' in a password field" doesnt convice me :/ 

(But agreed that my current PR is a bit dirty and hackish)

alexAubin:
Hmokay let's forget about this :sweat_smile: I must have missed something but it seems possible to pass booleans during GET requests and DELETE requests so must be possible somehow to do it in a POST request...

Will reopen if needed

PR #171, opened 2018-10-02
Protect against CSRF

randomstuff:


randomstuff:
Tests are passing. Still gotta check if it's actually working. I have to figure out how to test this properly.

randomstuff:
This strategy probably won't work happily with `Access-Control-Allow-Origin`. Do you actually need this?

randomstuff:
I think `Access-Control-Allow-Origin` is useless because `Access-Control-Allow-Methods` is not set. It shoul dprobably be removed.

zamentur:
Thanks for your contribution :)

Note: in general case the web admin on the same domain use this API. BUT some advanced users could use an external tools to call this API.

So we need to make a choice about CORS, restrict to the general case or be flexible.

randomstuff:
I udpates the code. This version only checks if `X-Requested-With` is set. This is simpler and should workd as a simple workaround.

randomstuff:
Note that this commit will break some applications such as [vpnclient_ynn](https://github.com/labriqueinternet/vpnclient_ynh/blob/55643202489c7336f3bb8a5c8912a07ab2e7dad8/conf/ynh-vpnclient-loadcubefile.sh#L99). Apparently most `labriqueinternet` apps use the same pattern and would break.

I guess they should use the CLI interface instead (?).

randomstuff:
> apparently it can be broken with some versions of Flash

AFAIU this is only true if you have `/crossdomain.xml` (and a Flash plugin installed).

alexAubin:
Hmmm I'm no CSRF expert, but reading the discussion, naively that looks ok/good enough ? Any opinion on this ?

moulinette/interfaces/api.py
```diff
@@ -27,6 +28,35 @@
...
+
+    if request.method != "POST":
+        return False
+    if request.content_type is None:
+        return True
+    content_type = request.content_type.lower().split(';')[0]
+    if content_type not in CSRF_TYPES:
+        return False
+
+    return request.headers.get("X-Requested-With") is None
```
alexAubin: So, trying to understand this, very naive question : what prevents the attacker's code from adding this header to the request ? Aren't CSRF attacks designed to target specific pieces of software and therefore the attacker would implement this header ? Am I missing something ?

randomstuff: The short answer is that the browser forbids the attacker's origin to set a custom HTTP Header unless the attacked origin explicitely allow that (with a `Access-Control-Allow-Headers` HTTP header).

The attack we are trying to prevent here (CSRF) is that the attacker code is executing from another origin (i.e. on another website). According to the fetch spec (https://fetch.spec.whatwg.org/), the attacker code is not allowed (by the browser) to add a custom HTTP Header (unless this is explicitely allowed by the target origin). It can only set `accept`, `accept-language`, `content-language`, `content-type` and few oddities.

alexAubin: Okay that sounds good to me then ! Thanks for the explanation !

PR #172, opened 2018-10-05
[enh] Migrate moulinette to python 3

zamentur:
## The problem

Python 2 will reach its end of support in 1 year.

## Solution

Convert moulinette into python3

## PR Status

WIP

## How to test

```
apt install python3-yaml python3-bottle python3-gnupg python3-argcomplete python3-psutil
rm -Rf /usr/lib/python3/dist-packages/moulinette; cp -R /vagrant/moulinette/moulinette  /usr/lib/python3/dist-packages/moulinette
cp /vagrant/moulinette/data/actionsmap/test.yml /usr/share/moulinette/actionsmap/ 
mkdir /usr/lib/moulinette/test
cp /vagrant/moulinette/lib/test/test.py /usr/lib/moulinette/test/ 
cp /vagrant/moulinette/bin/test-moulinette /usr/bin/test-moulinette
chmod u+x /usr/bin/test-moulinette

# Install ldap python3 module
apt install python3-pyldap

test-moulinette test non-auth
```

## Validation

- [ ] Principle agreement 0/2 : 
- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ dist
 build
 eggs
 parts
-bin
 cache
 var
 sdist
--- a/bin/test-moulinette
+++ b/bin/test-moulinette
@@ -1,4 +1,4 @@
-#! /usr/bin/python
+#! /usr/bin/python3
 # -*- coding: utf-8 -*-
 
 import os
@@ -163,7 +163,7 @@ def _init_moulinette(debug=False, quiet=False):
     # Create log directory
     if not os.path.isdir(LOG_DIR):
         try:
-            os.makedirs(LOG_DIR, 0750)
+            os.makedirs(LOG_DIR, 0o750)
         except os.error as e:
             _die(str(e))
 
--- a/bin/test-moulinette
+++ b/bin/test-moulinette
@@ -1,0 +1,215 @@
+#! /usr/bin/python
+# -*- coding: utf-8 -*-
+
+import os
+import sys
+import argparse
+
+# Either we are in a development environment or not
+IN_DEVEL = False
+
+# Level for which loggers will log
+LOGGERS_LEVEL = 'DEBUG'
+TTY_LOG_LEVEL = 'INFO'
+
+# Handlers that will be used by loggers
+#  - file: log to the file LOG_DIR/LOG_FILE
+#  - tty: log to current tty
+LOGGERS_HANDLERS = ['file', 'tty']
+
+# Directory and file to be used by logging
+LOG_DIR = '/var/log/yunohost'
+LOG_FILE = 'yunohost-cli.log'
+
+# Check and load - as needed - development environment
+if not __file__.startswith('/usr/'):
+    IN_DEVEL = True
+if IN_DEVEL:
+    basedir = os.path.abspath('%s/../' % os.path.dirname(__file__))
+    if os.path.isdir(os.path.join(basedir, 'moulinette')):
+        sys.path.insert(0, basedir)
+    LOG_DIR = os.path.join(basedir, 'log')
+
+
+import moulinette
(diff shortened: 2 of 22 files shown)


zamentur:
> root@yunohost:~# caniusepython3 --projects /vagrant/moulinette/moulinette
> Finding and checking dependencies ...
> [WARNING] problem fetching /vagrant/moulinette/moulinette, assuming ported (404)
> 
> 🎉  You have 0 projects blocking you from using Python 3!

zamentur:
Is this dependencies available in python 3 ?

- [x] python-ldap,
- [x] python-yaml,
- [x] python-bottle (>= 0.12),
- [x] python-gnupg,
- [x] python-gevent-websocket,
- [x] python-argcomplete,
- [x] python-psutil

alexAubin:
ytkwrkk

alexAubin:
Bump...

decentral1se:
What's the status on this @zamentur? If you don't have time to work on it, should I try to continue this work? Otherwise, I am glad to support in some way. Thanks for this effort :+1:

zamentur:
Feel free to contribute on this PR. I think it should be a team work, because it's a big big work.

decentral1se:
The thing about this PR is that it seems to have a lot of new testing changes? I mean, I want tests too but as far as I understand that is for another large piece of work. There is also no clear strategy on unit testing vs. functional testing vs. letting users smoke things out, as far as I understand. Could we reduce the work of this by sticking *only* to the py2 -> py3 changes? It's risky for users but more realistic plan for the lack of hands here on the development side.

Josue-T:
> The thing about this PR is that it seems to have a lot of new testing changes? I mean, I want tests too but as far as I understand that is for another large piece of work. There is also no clear strategy on unit testing vs. functional testing vs. letting users smoke things out, as far as I understand. Could we reduce the work of this by sticking _only_ to the py2 -> py3 changes? It's risky for users but a more realistic plan for the lack of hands here on the development side.

My point of view is that it could be really good to complete [these](https://github.com/YunoHost/yunohost/tree/stretch-unstable/src/yunohost/tests) unit test, to be able to test everything when we migrate to python3. By example here there are nothing which test the user management and for the app not a lot of things.

Working on the unit test wont fix this PR but it might be really helpfull in long term.

decentral1se:
From my limited experience and discussion on #yunohost-dev, getting a reliable test suite into place is something that has always been somehow out of reach. It's a massive effort. I think for the case of getting Python 3 out (hard deadline coming fast), we should stick to as small as possible py2 -> py3 diffs and rely on our own and user manual early testing and feedback on pre-releases. Isn't that more or less the standard way of doing this right now?

Open to ideas here since I am fairly new to this project. Just want to get a realistic plan.

alexAubin:
Uyea idk, from my understanding those tests were added because the changes were spread in the whole code that some way was needed to validate that nothing was broken ... Ideally having both unit/functional tests and "human" tests would be nice ?

But anyway right now the largest remaining work is to also port the yunohost core to python 3 ...

(Dunno if that addresses the question)

decentral1se:
OK, it's kinda clear to me now that we'll just get into a mess if we try to migrate `yunohost/yunohost` and `yunohost/moulinette` without some testing infrastructure. So I've made a start at that in https://github.com/YunoHost/moulinette/pull/203. Please let me know what you think!

decentral1se:
Coming back to this via https://github.com/YunoHost/moulinette/pull/225 :rocket: 

Let's not delete this branch btw, so that we can take from it if we need it later.

PR #173, opened 2018-10-09
[fix] Unapropriate warning

zamentur:
https://forum.yunohost.org/t/warning-moulinette-authenticator-ldap/5189/5

### DIFF
--- a/moulinette/authenticators/ldap.py
+++ b/moulinette/authenticators/ldap.py
@@ -59,6 +59,8 @@ class Authenticator(BaseAuthenticator):
 
     @property
     def is_authenticated(self):
+        if not self.con:
+            return False
         try:
             # Retrieve identity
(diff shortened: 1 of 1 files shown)


Psycojoker:
> Not sure why this didn't happen before stretch tho o.O

Looks like it's when I've added warning message everywhere because the previous code strategy was to hide all error messages.

alexAubin:
Sooo, idk, merging soonish ?

moulinette/authenticators/ldap.py
```diff
@@ -59,6 +59,8 @@ def __del__(self):
 
     @property
     def is_authenticated(self):
+        if not self.con:
```
Psycojoker: The code should be:

```suggestion
        if self.con is None:
```

As travis is complaining about (do you read thoses?)

PR #174, opened 2018-10-11
Remove Access-Control-Allow-Origin

randomstuff:
I don't think it's any useful without Access-Control-Allow-Methods anyway.


alexAubin:
Anybody has an opinion on this ?

PR #178, opened 2018-11-22
Fix coding style

airwoodix:
## Problem

The Travis-CI build fails for coding style reasons (first occurence quickly found is build [#278](https://travis-ci.org/YunoHost/moulinette/builds/419280251)). Some modules import unnecessary symbols.

## Proposed solution

Fix coding style according to PEP8. Remove unused imports except for those explicitly marked as voluntary.

## Status

Changes to make the CI build work are trivial and in principle harmless. Removal of unused imports may break client code (that shouldn't be) relying on them. Didn't check thoroughly for module import side effects within the `moulinette` code itself but doesn't seem to be the case. Ready for review.

### DIFF
--- a/moulinette/actionsmap.py
+++ b/moulinette/actionsmap.py
@@ -230,6 +230,7 @@ class RequiredParameter(_ExtraParameter):
                             % value)
         return value
 
+
 """
 The list of available extra parameters classes. It will keep to this list
 order on argument parsing.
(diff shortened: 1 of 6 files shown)


alexAubin:
Made a quick review, looks legit to me as well so merging as a micro decision.

Thanks again !

PR #180, opened 2018-11-25
[enh] Simplify moulinette errors

irina11y:
The problem

Utilisation trop complexe. Fait perdre du temps de dev.
Solution

Supression des code d'erreur "errno et de l'internationalisation "m18n.n"

this PR depends of  https://github.com/YunoHost/yunohost/pull/574

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -27,7 +27,6 @@
     "operation_interrupted": "Operation interrupted",
     "password": "Password",
     "pattern_not_match": "Does not match pattern",
-    "permission_denied": "Permission denied",
     "root_required": "You must be root to perform this action",
     "server_already_running": "A server is already running on that port",
     "success": "Success!",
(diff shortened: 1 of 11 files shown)


alexAubin:
This should now be ready for final review (and careful final re-reading to spot any typo that would have been missed so far).

(The branch has been rebased to current stretch-unstable and misc fixes have been implemented, tested the whole thing with the unit/functional tests)

alexAubin:
Meh I'm discovering that in `api.py` there is a function `error_to_reponse` which converts the (old) errno to an HTTP error code : 40x or 500. We gotta see what to do about this but in the meantime, this PR probably breaks the webadmin.

alexAubin:
- [X] replaced `StandardError` by `Exception` (though that was to be done in https://github.com/YunoHost/moulinette/pull/187 anyway)

About `error_to_response`, to me it looks like we should just remove it and trigger Error 400 in ~all cases. Also see this issue/comment in yunohost admin which is related : https://github.com/YunoHost/yunohost-admin/pull/220/#issuecomment-447373056

PR #181, opened 2018-11-26
Add possiblity to get attribute name of conflict in LDAP

Josue-T:
# Problem

- When we call the function `validate_uniqueness` we can't know if there are a conflict we don't know which attribute have generate a conflict and with which object we have a conflict.
- We can't manage the error with the function which call `validate_uniqueness`. Result is that if there are a conflict the user just see a message "Attribute '{attribute}' already exists with value '{value}'" but the user might not know the meaning of the attribute name.

# Solution

- Create a function `get_conflict` which just return the name of the attribute and it value if there are a conflict.

### DIFF
--- a/moulinette/authenticators/ldap.py
+++ b/moulinette/authenticators/ldap.py
@@ -260,6 +260,4 @@ class Authenticator(BaseAuthenticator):
                 continue
             else:
                 return (attr, value)
-                logger.info("attribute '%s' with value '%s' is not unique",
-                            attr, value)
         return None
--- a/moulinette/authenticators/ldap.py
+++ b/moulinette/authenticators/ldap.py
@@ -234,13 +234,31 @@ class Authenticator(BaseAuthenticator):
             Boolean | MoulinetteError
 
         """
+        attr_found = self.get_conflict(value_dict)
+        if attr_found:
+            logger.info("attribute '%s' with value '%s' is not unique",
+                        attr_found[0], attr_found[1])
+            raise MoulinetteError(errno.EEXIST,
+                                    m18n.g('ldap_attribute_already_exists',
+                                            attribute=attr_found[0], value=attr_found[1]))
+        return True
+
+    def get_conflict(self, value_dict, base_dn=None):
+        """
+        Check uniqueness of values
+
+        Keyword arguments:
+            value_dict -- Dictionnary of attributes/values to check
+
+        Returns:
+            None | list with Fist conflict attribute name and value
+
+        """
         for attr, value in value_dict.items():
-            if not self.search(filter=attr + '=' + value):
+            if not self.search(base=base_dn, filter=attr + '=' + value):
                 continue
             else:
+                return (attr, value)
                 logger.info("attribute '%s' with value '%s' is not unique",
                             attr, value)
-                raise MoulinetteError(errno.EEXIST,
-                                      m18n.g('ldap_attribute_already_exists',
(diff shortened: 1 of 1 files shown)


moulinette/authenticators/ldap.py
```diff
@@ -233,14 +233,33 @@ def validate_uniqueness(self, value_dict):
...
+        if attr_found:
+            logger.info("attribute '%s' with value '%s' is not unique",
+                        attr_found[0], attr_found[1])
+            raise MoulinetteError(errno.EEXIST,
+                                  m18n.g('ldap_attribute_already_exists',
+                                         attribute=attr_found[0],
+                                         value=attr_found[1]))
+        return True
+
+    def get_conflict(self, value_dict, base_dn=None):
```
alexAubin: What is the motivation in passing `base_dn` in args if it's not used yet ? Are there foreseen usage of this ?

moulinette/authenticators/ldap.py
```diff
@@ -233,14 +233,33 @@ def validate_uniqueness(self, value_dict):
...
+            None | list with Fist conflict attribute name and value
+
         """
         for attr, value in value_dict.items():
-            if not self.search(filter=attr + '=' + value):
+            if not self.search(base=base_dn, filter=attr + '=' + value):
                 continue
             else:
+                return (attr, value)
                 logger.info("attribute '%s' with value '%s' is not unique",
```
alexAubin: Shouldnt this logger.info be removed if there's a return right before it ?

moulinette/authenticators/ldap.py
```diff
@@ -233,14 +233,33 @@ def validate_uniqueness(self, value_dict):
...
 
+        """
+        attr_found = self.get_conflict(value_dict)
+        if attr_found:
+            logger.info("attribute '%s' with value '%s' is not unique",
+                        attr_found[0], attr_found[1])
+            raise MoulinetteError(errno.EEXIST,
+                                  m18n.g('ldap_attribute_already_exists',
+                                         attribute=attr_found[0],
+                                         value=attr_found[1]))
```
alexAubin: Uh in fact I don't understand how that changes the behavior compared to the current code o.O ? Am i missing something o.O ?

moulinette/authenticators/ldap.py
Josue-T: Because sometime I don't want all element in all part in the LDAP tree. By example want only to now in the group subtree  if there are a conflict.

moulinette/authenticators/ldap.py
Josue-T: I just adapted the function "validate_uniqueness" to don't duplique the code. I kept this function to keep the compatibility with the actual code in the core

moulinette/authenticators/ldap.py
alexAubin: Okay I didn't realise this was a method for the auth object and that you are using this in https://github.com/YunoHost/yunohost/pull/585

moulinette/authenticators/ldap.py
alexAubin: Okay so in fact now i understand the motivation of this PR which is (if i understood correctly) to factorize the get_conflict part so that you can use it in YunoHost core like in YunoHost/yunohost#585

Josue-T: Yes :+1:

PR #182, opened 2018-11-26
Make sure `gpg.encrypt` actually does something

alexAubin:
So, today somebody encountered this stupid issue where he could not login anymore to the webadmin.

Turns out that for some reasons moulinette uses gpg to encrypt some values in cache. The issue lies in the fact that if there's an error in `/root/.gnupg/gpg.conf` (such as a bad option), `gpg.encrypt()` will simply return an empty string...

With verbosity on (`gnupg.GPG(verbose=True)`), we were able to see : `gpg: /root/.gnupg/gpg.conf:2: invalid`. The user told me that he was using duply for backups which also relies on gpg, and that might be related.

Anyway, I propose to add an assertion to check this to be able to pinpoint the issue more easily when this happens...

### DIFF
--- a/moulinette/authenticators/__init__.py
+++ b/moulinette/authenticators/__init__.py
@@ -138,9 +138,13 @@ class BaseAuthenticator(object):
         """Store a session and its associated password"""
         gpg = gnupg.GPG()
         gpg.encoding = 'utf-8'
+
+        # Encrypt the password using the session hash
+        s = str(gpg.encrypt(password, None, symmetric=True, passphrase=session_hash))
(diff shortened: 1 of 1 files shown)


randomstuff:
Would it make sense to use itsdangerous for this kind of thing, instead?

alexAubin:
Uuuuh, I dunno, I guess the purpose here is to encrypt the data ? I didnt have a deep look in what is done and how it's been designed, this is just a "bugfix" to make it easier to debug next time this happens ...

PR #183, opened 2018-11-29
Use sasl authentication for LDAP (by root user)

Josue-T:
# Problem

- For each action in LDAP we need to give the admin password.
- Will be a problem if we need to update LDAP in the app install scripts.

# Solutions

Use sasl authentication for LDAP. By that LDAP authenticate the user by his UID. So if the user is root it will be automatically authenticated.

Note that this PR depends reciprocally of https://github.com/YunoHost/yunohost/pull/585

The main important change in the yunohost repos is this change in the actionmap : 
```diff
            default:
                vendor: ldap
                help: admin_password
                parameters:
                    uri: ldap://localhost:389
                    base_dn: dc=yunohost,dc=org
-                    user_rdn: cn=admin
+                    user_rdn: cn=admin,dc=yunohost,dc=org
```
# How to test

With ynhdev :

- for the yunohost repos checkout the branch "group_permission"
- For the moulinette repos checkout the branch "LDAP_validate_uniqueness" and "sasl_authentication"
- Do the postinstall
- Do any action which need LDAP (create user, create group, update permission, etc) and see that you don't need any password.

## PR Status

Ready for review.


alexAubin:
Mergin', c.f. https://github.com/YunoHost/yunohost/pull/721#issuecomment-494862010

PR #184, opened 2018-12-09
[enh] Display date as system timezone

zamentur:
## The problem
Dates are not well displayed. Currently it's just a str(date) and in some case yunohost give a timestamp... We need to harmonize that, and manage timezone.

## Solution

Display dates using system timezone if we use cli.

Note: With the api, the json encoder already use iso format on the date object, so if the date is not "naive" date the representation generated by the api contains timezone info.
https://github.com/YunoHost/moulinette/blob/stretch-unstable/moulinette/utils/serialize.py#L30

## Status
Ready
This PR is linked to 
https://github.com/YunoHost/yunohost/pull/552

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -16,7 +16,8 @@ Depends: ${misc:Depends}, ${python:Depends},
     python-gnupg,
     python-gevent-websocket,
     python-argcomplete,
-    python-psutil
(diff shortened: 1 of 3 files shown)


zamentur:
Ready and tested

PR #185, opened 2018-12-11
Fix LDAP authenticator when slapd was restarted

Josue-T:
# Problem

When we restart slapd the authenticator object is broken. The result of that is that we need to make some hack like [that](https://github.com/YunoHost/yunohost/blob/stretch-unstable/src/yunohost/backup.py#L897-L899).

```
            # Quickfix: the old app_ssowatconf(auth) instruction failed due to
            # ldap restore hooks
            os.system('sudo yunohost app ssowatconf')
```

This issue also impact all test which are made after the backup test.

# Solution

Use an authenticator which are able to make a new connection if the connection is lost.

### DIFF
--- a/moulinette/authenticators/ldap.py
+++ b/moulinette/authenticators/ldap.py
@@ -76,7 +76,7 @@ class Authenticator(BaseAuthenticator):
 
     def authenticate(self, password):
         try:
-            con = ldap.initialize(self.uri)
+            con = ldap.ldapobject.ReconnectLDAPObject(self.uri, retry_max=5, retry_delay=0.1)
             if self.userdn:
                 con.simple_bind_s(self.userdn, password)
             else:


moulinette/authenticators/ldap.py
```diff
@@ -76,7 +76,7 @@ def is_authenticated(self):
 
     def authenticate(self, password):
         try:
-            con = ldap.initialize(self.uri)
+            con = ldap.ldapobject.ReconnectLDAPObject(self.uri, retry_max=5, retry_delay=0.1)
```
alexAubin: 0.1 sec for `retry_delay` sounds a bit short ? What about 0.5 ?

Josue-T: Well, maybe Yes.

I need to do some more test. I firstly tried with 10 (which is clearly too long). And the test failed because the connection was not re-establish yet.

So this parameter might depends of the processor. If the CPU is really quick it might be not too short.

Josue-T: Tested with 3 sec and it's also OK.

I purpose to put `retry_max=10, retry_delay=0.5`

PR #186, opened 2018-12-12
Create possiblity to add an preaction

Josue-T:
# Problem

I need to be able to do some action before to call the main action in the moulinette.

It's used by the PR : https://github.com/YunoHost/yunohost/pull/599

# Solution

Add a generic concept of action which can be called before the main action.

The actionmap is used to define what will be do. By example : 
```
    preaction:
        yunohost.preactions.checkmigrations:
            - main
```
Mean that the module function `main` from the module `yunohost.preactions.checkmigrations` will be executed.

The function main have two arguments:
- The  first argument is the function name of the Yunohost operation which will be done.
- The second will a dictionnary of the argument passed to the Yunohost operation.

### DIFF
--- a/moulinette/actionsmap.py
+++ b/moulinette/actionsmap.py
@@ -490,6 +490,32 @@ class ActionsMap(object):
 
         # Lock the moulinette for the namespace
         with MoulinetteLock(namespace, timeout):
+            # Load and execute preactions
+            if self.parser.get_preactions():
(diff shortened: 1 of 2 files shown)


alexAubin:
Sooo, following discussions about the way to handle this, the migration mechanism should be changed as proposed in another PR instead of locking half of yunohost :s

PR #201, opened 2019-05-25
Create a LDIF parser

Josue-T:
Create a parser for the LDIF file.

I need this for https://github.com/YunoHost/yunohost/pull/585


alexAubin:
Yolomergin because looks legit and is a standalone change, independent from other stuff

PR #203, opened 2019-06-28
Python3 migration prep: Tox, Pytest and migrating existing unit tests

decentral1se:
This PR tries to setup some testing infrastructure in preparation for migrating to Python 3 using best-in-class tools (Tox, Pytest, Pep8, etc.). After the discussion in https://github.com/YunoHost/moulinette/pull/172 and after looking into `moulinette` a bit more, I'm now thinking we need to work on the tests first. So, this is my first step.

See https://travis-ci.org/YunoHost/moulinette/builds/551898217 for the Tox build and coverage report. We're at 11% already :)


decentral1se:
Any thoughts on this?

Psycojoker:
Link to https://github.com/YunoHost/moulinette/pull/172

decentral1se:
I haven't seen much opposition to this? If this can move forward then I will continue to work on increasing the test coverage of moulinette and see what progress I can make. We'll then be in a good position to use Tox (https://tox.readthedocs.io/en/latest/#what-is-tox) to test the `py35`, `py36`, `py37`, etc. environments against the `2to3` migrated code base and have some idea of confidence for the upgrade.

Psycojoker:
> I haven't seen much opposition to this? 

Yeah, I think you can go forward :) Just be aware that we are all, like, super busy on so many things, so it's kinda normal to not have a response fast :( (yeah it sucks)

decentral1se:
Cool! 

I would like to merge this then now as is so that the next diff will not be huge.

It will be easier to review in the following steps.

?

moulinette/utils/filesystem.py
```diff
@@ -31,8 +31,8 @@ def read_file(file_path):
             file_content = f.read()
     except IOError as e:
         raise MoulinetteError('cannot_open_file', file=file_path, error=str(e))
-    except Exception as e:
-        raise MoulinetteError('error_reading_file', file=file_path, error=str(e))
+    except Exception:
+        raise MoulinetteError('unknown_error_reading_file', file=file_path)
```
alexAubin: Eh they key did not exist ? Note that this mistake about this key is also later around line 108 ...

Ideally it would be nice to keep the `error=str(e)` in the final error message though I see that it's not used in the i18n strings so that's probably something to be careful about :/ ...

test/conftest.py
```diff
@@ -0,0 +1,128 @@
...
+    test_file.write_bytes(test_text)
+    return test_file
+
+
+@pytest.fixture
+def test_json(tmp_path):
+    test_json = json.dumps({'foo': 'bar'})
+    test_file = tmp_path / 'test.json'
+    test_file.write_bytes(test_json)
+    return test_file
```
alexAubin: :warning: (Post scriptum : Below is probably a very ignorant and pytest-noob remark, sorry about that)

Uh I'm not sure to understand what this is about ...? (Maybe this is from an old file copied, didn't check, sorry)

I don't understand how a path like `tmp_path` (supposedly a string ?) can be *divided* by a file name ... and then run `write_bytes` on the result ?

Also `test_json` doesn't seem to be used anywhere ?

Or is it some pytest fixture magic ? (Never really tried to understood those) But even then, what's `tmp_path` ?

alexAubin:
(well maybe modulo the small thing about `unknown_error_reading_file`, not sure what to do)

test/conftest.py
decentral1se: Pytest fixtures are amazing. But a total pain in the ass at the start. Take a quick look at the top section of https://docs.pytest.org/en/latest/tmpdir.html and you'll see what this all means. `tmp_path` is a https://github.com/mcmtroffaes/pathlib2 instance. `test_json` gets written to the temporary file.

moulinette/utils/filesystem.py
decentral1se: Yep, it did not exist. Or, I only found `unknown_error_reading_file` in the locales files. I've added the `{error:s]` to the i18n string and also fixed the usages ~ line 108. That's done in the 75c8517cdef6d74e3fa3aa25ba1a4456b4cf3314 fixup commit (will rebase it all once the reviews are done).

decentral1se:
Pushed a small fixup and tried to clear up the point about pytest fixtures.

test/conftest.py
alexAubin: Thanks ! I don't have time right now, but will definitely read this documentation, that sounds really worth reading ;)

alexAubin:
Yolomerging then ! Thanks a lot again !

decentral1se:
Linking to https://github.com/YunoHost/issues/issues/1255 so people can see what progress is being made for the Buster related Python 3 migration.

PR #204, opened 2019-07-07
add utils.filesystem.read_toml and depends on toml in moulinette

Psycojoker:
Pretty much straightforward

### DIFF
--- a/moulinette/utils/filesystem.py
+++ b/moulinette/utils/filesystem.py
@@ -91,7 +91,7 @@ def read_toml(file_path):
     # Read file
     file_content = read_file(file_path)
 
-    # Try to load toml to check if it's syntaxically correct
+    # Try to load toml to check if it's syntactically correct
     try:
         loaded_toml = toml.loads(file_content, _dict=OrderedDict)
     except Exception as e:
--- a/moulinette/utils/filesystem.py
+++ b/moulinette/utils/filesystem.py
@@ -1,10 +1,13 @@
 import os
 import yaml
+import toml
 import errno
 import shutil
 import json
 import grp
+
 from pwd import getpwnam
+from collections import OrderedDict
 
 from moulinette import m18n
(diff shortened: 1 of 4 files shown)


moulinette/utils/filesystem.py
```diff
@@ -77,6 +80,28 @@ def read_yaml(file_path):
...
+    Safely read a toml file
+
+    Keyword argument:
+        file_path -- Path to the toml file
+    """
+
+    # Read file
+    file_content = read_file(file_path)
+
+    # Try to load toml to check if it's syntaxically correct
```
decentral1se:
```suggestion
    # Try to load toml to check if it's syntactically correct
```

Unsure if "syntaxically" is a new cool way to say this though :)

moulinette/utils/filesystem.py
```diff
@@ -77,6 +80,28 @@ def read_yaml(file_path):
...
+    try:
+        loaded_toml = toml.loads(file_content, _dict=OrderedDict)
+    except Exception as e:
+        raise MoulinetteError(errno.EINVAL,
+                              m18n.g('corrupted_toml',
```
decentral1se: We're missing the `corrupted_toml` in the locales files?

moulinette/utils/filesystem.py
kay0u: In french we have "syntaxique" for "syntactic", like "connexion" for "connection" and a few others like that, it's very confusing --''.

moulinette/utils/filesystem.py
Psycojoker: Ah yes indeed '^'

Psycojoker:
I realize that this PR is missing for the release, I'm surprise this isn't crashing the app code '^'

alexAubin:
Yolomergin' ¯\\_(ツ)\_/¯

PR #205, opened 2019-07-20
Warn the user about long locks for a way better UX

Psycojoker:
Today (and like in several other cases) I end up in front of a user convinced that their YunoHost was totally broken because they though that nothing worked anymore. I've suspected the lock directly but it was hard to debug because even `--debug` didn't informed that YunoHost was waiting for the lock.

So this PR add 2 mechanism to fix this:

* add a simple `logger.debug` line to tell that we are entering the lock mechanism
* add a gradual `logger.warning` mechanism to tell the user that we are waiting and still waiting for the other command to finish

My formulation for the second error message might not be very clear so if you have a better suggestion try to improve it please.

I'm also expecting those message to bring users try to understand what is happening and how to debug that, one better solution for that would be to have a specific documentation page for this situation and link to it, but we don't have one :/ So in the meantime this already seems like a needed improvement.

Another way way better improvement would be for every command that take to lock to say what it is doing (like "listing domain") and that's really an improvement I'm considering doing and we could use the "description" already available in the `actionmaps.yaml` or add this information in it. But that's much more work that I can do right now :/

On "what would be the so much better solution that we won't have the time to do, lol": I really think that YunoHost should swtich from a lock oriented approach to a "jobs list" approach where you can see at anytime all the current process that are waiting and running and killing or reordering all of the them. But that's really way much more work.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -53,5 +53,6 @@
     "download_bad_status_code": "{url:s} returned status code {code:s}",
     "command_unknown": "Command '{command:s}' unknown ?",
     "warn_the_user_about_waiting_lock": "Another YunoHost command is running right now, we are waiting for it to finish before running this one",
-    "warn_the_user_about_waiting_lock_again": "Still waiting..."
+    "warn_the_user_about_waiting_lock_again": "Still waiting...",
+    "warn_the_user_that_lock_is_acquired": "the other command just complet, now starting this command"
 }
--- a/locales/en.json
+++ b/locales/en.json
@@ -51,5 +51,7 @@
     "download_timeout": "{url:s} took too long to answer, gave up.",
     "download_unknown_error": "Error when downloading data from {url:s}: {error:s}",
     "download_bad_status_code": "{url:s} returned status code {code:s}",
-    "command_unknown": "Command '{command:s}' unknown?"
+    "command_unknown": "Command '{command:s}' unknown ?",
+    "warn_the_user_about_waiting_lock": "another YunoHost command is running right now, we are waiting for it to finish before running this one",
+    "warn_the_user_about_waiting_lock_again": "still waiting..."
 }
--- a/locales/en.json
+++ b/locales/en.json
@@ -52,6 +52,6 @@
     "download_unknown_error": "Error when downloading data from {url:s}: {error:s}",
     "download_bad_status_code": "{url:s} returned status code {code:s}",
     "command_unknown": "Command '{command:s}' unknown ?",
-    "warn_the_user_about_waiting_lock": "another YunoHost command is running right now, we are waiting for it to finish before running this one",
+    "warn_the_user_about_waiting_lock": "Another YunoHost command is running right now, we are waiting for it to finish before running this one",
     "warn_the_user_about_waiting_lock_again": "Still waiting..."
 }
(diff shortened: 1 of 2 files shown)


locales/en.json
```diff
@@ -51,5 +51,7 @@
     "download_timeout": "{url:s} took too long to answer, gave up.",
     "download_unknown_error": "Error when downloading data from {url:s}: {error:s}",
     "download_bad_status_code": "{url:s} returned status code {code:s}",
-    "command_unknown": "Command '{command:s}' unknown?"
+    "command_unknown": "Command '{command:s}' unknown ?",
+    "warn_the_user_about_waiting_lock": "another YunoHost command is running right now, we are waiting for it to finish before running this one",
```
decentral1se:
```suggestion
    "warn_the_user_about_waiting_lock": "Another YunoHost command is running right now, we are waiting for it to finish before running this one",
```

locales/en.json
```diff
@@ -51,5 +51,7 @@
...
+    "warn_the_user_about_waiting_lock_again": "still waiting..."
```
decentral1se:
```suggestion
    "warn_the_user_about_waiting_lock_again": "Still waiting..."
```

decentral1se:
Oh, btw, it might actually be possible to write a test for this now! Please see the usage of the `caplog` fixture with Pytest in https://docs.pytest.org/en/latest/logging.html#caplog-fixture. You just add `caplog` as an argument to your test, run the function and then iterate over the captured logging and assert what you expect. The testing setup is still quite "experimental" (see #207) but just an FYI all the same :)

Psycojoker:
I haven't read the test (sorry '^') but does it really supports starting the whole moulinette mechanism with the full lock and all '-' ?

locales/en.json
```diff
@@ -51,5 +51,7 @@
     "download_timeout": "{url:s} took too long to answer, gave up.",
     "download_unknown_error": "Error when downloading data from {url:s}: {error:s}",
     "download_bad_status_code": "{url:s} returned status code {code:s}",
-    "command_unknown": "Command '{command:s}' unknown?"
+    "command_unknown": "Command '{command:s}' unknown ?",
+    "warn_the_user_about_waiting_lock": "another YunoHost command is running right now, we are waiting for it to finish before running this one",
+    "warn_the_user_about_waiting_lock_again": "Still waiting..."
```
alexAubin: https://www.youtube.com/watch?v=Kjwlzs1OJcY :+1:

decentral1se:
I am not sure for this case, so I just mentioned that it was "experimental" :) No worries for now and if you didn't get time, I know you are busy with other things. That's all fine. It's good enough to know that this kind of thing will be testable soon. For avoiding slowness, you probably want to just mock that the time calls are done. You also have https://github.com/ktosiek/pytest-freezegun.

Psycojoker:
Tests aren't here yet but I've added MOAR UX stuff.

Psycojoker:
After quite some experience using mock/patch @work and re-reading this code and thinking quite a bit about it I couldn't find a way to write tests for it that wouldn't be crazily mocking patching everything up to a point where I don't think it would make much sens anymore, this is way to to heavily side effect based :x (but I might be mistaking) 

So I'm merging it now, if you manage to get a prototype on how to write a test for that that make sens we can do it, but I withdraw from this attempt for now, sorry :x

PR #207, opened 2019-07-20
Refactor global loading

decentral1se:
This follows the work on #203.

So, while trying to write more tests, I am running into the issue where I cannot mock the `moulinette.globals` configuration values. This makes it hard to control the environment for the unit tests more closely. This is because they are imported at package setup, initialised at package load and also imported globally at each module level.

So, I have tried in this change to refactor how these values are exposed. For consuming this library, I've introduced a new API, `moulinette.env` which allows you to do `moulinette.env['CACHE_DIR]` instead of `from moulinette import CACHE_DIR`. This is a breaking change but I haven't found any code in yunohost/yunohost that relies on these properties being exposed. Is there somewhere to look? Otherwise, internally, I've tried to maintain backwards compatibility in all other places in the code base (for example, the `ActionsMap` class).

I've added a test, `test_open_cachefile_creates`, which shows what I have been trying to achieve all along. That we can easily just override the environment variables that Moulinette expects.

This change will make it much easier to test moulinette.

### DIFF
--- a/moulinette/__init__.py
+++ b/moulinette/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from moulinette.core import init_interface, MoulinetteError, MoulinetteSignals, Moulinette18n
-from moulinette.globals import DATA_DIR, LIB_DIR, LOCALES_DIR, CACHE_DIR
+from moulinette.globals import init_moulinette_env
 
 __title__ = 'moulinette'
 __version__ = '0.1'
@@ -28,9 +28,8 @@ __credits__ = """
     along with this program; if not, see http://www.gnu.org/licenses
     """
 __all__ = [
-    'init', 'api', 'cli', 'm18n',
+    'init', 'api', 'cli', 'm18n', 'env',
     'init_interface', 'MoulinetteError',
-    'DATA_DIR', 'LIB_DIR', 'LOCALES_DIR', 'CACHE_DIR',
 ]
 
 
@@ -62,7 +61,7 @@ def init(logging_config=None, **kwargs):
     configure_logging(logging_config)
 
     # Add library directory to python path
-    sys.path.insert(0, LIB_DIR)
+    sys.path.insert(0, init_moulinette_env()['LIB_DIR'])
 
 
 # Easy access to interfaces
@@ -139,3 +138,8 @@ def cli(namespaces, args, use_cache=True, output_as=None,
         logging.getLogger(namespaces[0]).error(e.strerror)
         return 1
     return 0
+
+
+def env():
+    """Initialise moulinette specific configuration."""
+    return init_moulinette_env()
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,10 @@
 import os
 import sys
 from setuptools import setup, find_packages
+from moulinette.globals import init_moulinette_env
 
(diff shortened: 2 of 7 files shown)


moulinette/__init__.py
```diff
@@ -62,7 +61,8 @@ def init(logging_config=None, **kwargs):
     configure_logging(logging_config)
 
     # Add library directory to python path
-    sys.path.insert(0, LIB_DIR)
+    # Note(decentral1se): must match moulinette.globals value
+    sys.path.insert(0, '/usr/lib/moulinette')
```
alexAubin: Uh so here are we really sure we want to hardcode this value here ?

decentral1se: OK, fair concern. I've taken another look and also used the `init_env` trick. It seems to work and the tests pass, so it looks like we can do this just fine. Please see 1854ab360bb32c17554f4394f63293685f42e700.

decentral1se:
OK, because of https://travis-ci.org/YunoHost/moulinette/builds/563903918#L265, I've had to add the https://github.com/YunoHost/moulinette/pull/207/commits/832c3692984f971d6af2606b5ee96f4760bbd6b5 commit to resolve that. The build passes now https://travis-ci.org/YunoHost/moulinette/builds/563905323.

setup.py
```diff
@@ -36,6 +38,7 @@
           'psutil',
           'pytz',
           'pyyaml',
+          'python-toml',
```
Psycojoker: Hu weird, `python-toml` is a debian pkg, I'm using `toml` from pypi, that's a mistake?

decentral1se: Aha, you mean https://pypi.org/project/toml/? I see that is done now in https://github.com/YunoHost/moulinette/commit/55515cb065f43d2bb7cee77504447f2b8924703a :+1: I'll drop this commit.

decentral1se:
https://travis-ci.org/YunoHost/moulinette/builds/564585440 :sun_with_face:

setup.py
Psycojoker: Yup, added it without realising that you did it and saw it conflicted on this PR :x

decentral1se:
* Rebased
* naming: init_env -> init_moulinette_env
* naming: env -> moulinette_env

moulinette/__init__.py
```diff
@@ -139,3 +138,8 @@ class at construction
         logging.getLogger(namespaces[0]).error(e.strerror)
         return 1
     return 0
+
+
+def env():
```
Psycojoker: Shouldn't it be "moulinette_env" there then? (and in `__all__`)

I'm not sure to understand the purpose of this function :x

decentral1se: For the naming: the rest of the API is quite "short", like "cli", "init", etc.

For the purpose: before you could `from moulinette import LOCALES_DIR` but that was causing problems (explained in first post). So now you can do `from moulinette import env` and then do `env['LOCALES_DIR']`.

Psycojoker: Oh ok, I guess we'll do some cleaning of that one day ™

Thx for the answer :)

decentral1se:
Linking to https://github.com/YunoHost/issues/issues/1255 so people can see what progress is being made for the Buster related Python 3 migration.

PR #208, opened 2019-07-23
Better debug information related to gpg for storing session

Psycojoker:
I'm trying to improve a bit this messy part about gnupg errors that more and more users start to get x(

But I'm don't have much time and it's quite not very intuitive extracted out of nowhere (and I'm not sure we do really need it anymore?)

At least this will give proper error messages.

### DIFF
--- a/moulinette/authenticators/__init__.py
+++ b/moulinette/authenticators/__init__.py
@@ -93,8 +93,8 @@ class BaseAuthenticator(object):
             try:
                 # Extract id and hash from token
                 s_id, s_hash = token
-            except TypeError:
-                logger.error("unable to extract token parts from '%s'", token)
+            except TypeError as e:
+                logger.error("unable to extract token parts from '%s' because '%s'", token, e)
                 if password is None:
                     raise MoulinetteError('error_see_log')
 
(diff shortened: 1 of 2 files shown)


Psycojoker:
We need to explore more but having read the code we need to ask the users that have this bug if they found this message in their log https://github.com/YunoHost/moulinette/blob/3c9230ee9f5676ed26700f96d0cdbda8cd7dd7fb/moulinette/authenticators/__init__.py#L101

That seems to be the only situation in which yunohost is happy at first but start failing later on for no reason x(

moulinette/authenticators/__init__.py
```diff
@@ -150,16 +152,16 @@ def _retrieve_session(self, session_id, session_hash):
...
         else:
             gpg = gnupg.GPG()
             gpg.encoding = 'utf-8'
 
             decrypted = gpg.decrypt(enc_pwd, passphrase=session_hash)
             if decrypted.ok is not True:
-                logger.error("unable to decrypt password for the session: %s",
-                             decrypted.status)
-                raise MoulinetteError('unable_retrieve_session')
+                error_message = "unable to decrypt password for the session: %s", decrypted.status
```
alexAubin: Eh, will this work ? Don't you need a `%` instead of the comma ?

Psycojoker: Ah, indeed x_x

PR #209, opened 2019-07-23
[mod] add gnupg1 as explicit dependency to avoid its removal

Psycojoker:
According to https://packages.debian.org/stretch/python-gnupg it's the "gnupg1" package that we need (and not "gnupg" itself) for some weird compatibility reasons.

Link to https://github.com/YunoHost/issues/issues/1326 in which apparently installing "gnupg2" pulled "gnupg" (but not gnupg1?) https://packages.debian.org/stretch/gnupg2 I'm not sure, it looks like a mess :/

Last time I've checked python-gnupg was using "gpg1" binary but the source code is annoying to read. Here is which pkg provide which binary:

```
root@ynh:/etc/yunohost/apps/ynhexample__2# apt-file search "bin/gpg"
gnupg: /usr/bin/gpg
gnupg: /usr/bin/gpg-zip
gnupg: /usr/bin/gpgconf
gnupg: /usr/bin/gpgparsemail
gnupg: /usr/bin/gpgsplit
gnupg-agent: /usr/bin/gpg-agent
gnupg-agent: /usr/bin/gpg-connect-agent
gnupg1: /usr/bin/gpg1
gnupg2: /usr/bin/gpg2

```

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -10,6 +10,7 @@ Homepage: https://github.com/YunoHost/moulinette
 Package: moulinette
 Architecture: all
 Depends: ${misc:Depends}, ${python:Depends},
+    gnupg1,
     python-ldap,
     python-yaml,
     python-bottle (>= 0.12),


alexAubin:
Eh, bug gnupg1 is a dependency of python-gnupg ...

```
 > aptitude why gnupg1
i   python-gnupg Depends gnupg1
```

```
 > apt remove gnupg1
[...]
The following packages will be REMOVED:
  gnupg1 gnupg1-curl moulinette python-gnupg yunohost yunohost-admin
WARNING: The following essential packages will be removed.
This should NOT be done unless you know exactly what you are doing!
  yunohost moulinette (due to yunohost)
0 upgraded, 0 newly installed, 6 to remove and 0 not upgraded.
After this operation, 20.9 MB disk space will be freed.
You are about to do something potentially harmful.
To continue type in the phrase 'Yes, do as I say!'
```

Psycojoker:
But how did it end up being removed and break moulinette then :/?

alexAubin:
Well uh I dunno, to be that's not so clear ... Are you really one hundred percent sure that the package wasn't installed anymore ? When I experienced this issue, as far as I remember, it was more like a very temporary issue because somehow "gpg was not ready to process the thing" (though not sure about it either ...)

But in any case I don't expect that adding this dependency in our control file will fix anything :/

Psycojoker:
Meh ok, I really don't know tbh :/

Maybe we should either just trash this gpg thing (I'm not even sure it's good from a security point of view) or wait for more debug information with this PR https://github.com/YunoHost/moulinette/pull/208

Psycojoker:
Closing then, waiting for better debugging situation.

PR #210, opened 2019-07-28
Add lint job to Tox and Travis

decentral1se:
So, let's use Tox to run pycodestyle (new name for pep8) and run it in a parallel job.

Why? Well, I want to also start to introduce new jobs to Tox / Travis.

In the end, the developer experience will always be "just run pip install tox && tox"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,14 @@
 language: python
-install: pip install tox pep8
-python:
-    - 2.7
-script: 
-  - pep8 moulinette
+
+matrix:
+  include:
+    - python: 2.7
+      env: TOXENV=py27
+    - python: 2.7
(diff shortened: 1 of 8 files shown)


decentral1se:
Look ma, two jobs, one tox.

https://travis-ci.org/YunoHost/moulinette/builds/564716895

Psycojoker:
Shouldn't we spawn flake8 directly? iirc it includes pycodestyle AND does some error checkings

decentral1se:
> Shouldn't we spawn flake8 directly?

I'm happy to use Flake8 as well! For others: http://flake8.pycqa.org/en/latest/.

Done in 4c79b0b164f9219f44c9c7158b5b13d05f454c41.

PR #216, opened 2019-08-20
Simplify auth mechanism

alexAubin:
Soooo I was crazy enough to try to investigate how to remove the dependency to `gnupg` which led me to try to understand the spaghetti madness of how/where the authentication happens exactly...

I ended up cleaning and refactorizing various parts of the code to make it less crazy ...

So right now what I got is this (trying to summarizing my understading .. that's still quite complicated but much less than before this PR I guess) : 

- each interface `api` and `cli` implement a `auth_required` method which dynamically check if the request requires to be authenticated. For example, in the context of yunohost, the special route `POST /postinstall` doesn't require to be authenticated
- `auth_required` is used by the `actionsmap.py`, which if needed loads and triggers the authentication check for the corresponding profile (in our case, basically we only have `default`...)
 - the loading of the authentication modules (well, `ldap`..) is done according to the details provided in the actionmap
 - then each interface `api` and `cli` implement a `_do_authenticate` method that allows to request some secret or whatever mechanism for the user to effectively authenticate
 - the actual authentication happens by calling `authenticator()` which calls `__call__` in `authenticators/__init__.py` itself calling `ldap.authenticate(password)`  (but then there's the whole token mechanism stored with gpg etc...)
 - finally, `ldap.is_authenticated` is called to check wether the user is or not authenticated now


alexAubin:
Sooo it's mostly done I think though I might still tweak a few minor things or add other tests ... so reviews are welcome if you have the courage for this. Reading commit by commit is encouraged.

@decentral1se : I'm interested if you have some feedback on how I wrote the tests .. I'm not sure that it's 100% the orthodox way of doing it. Also it's a bit tricky because I have to start the api thread in the setup - but then the coverage can't easily be computed. Maybe you know a trick ... (And last : dunno if there's a tricky to be able to run pdb when running tox - or running single test files manually)

moulinette/authenticators/__init__.py
```diff
@@ -135,33 +144,46 @@ def _open_sessionfile(self, session_id, mode='r'):
...
-            if decrypted.ok is not True:
-                error_message = "unable to decrypt password for the session: %s" % decrypted.status
-                logger.error(error_message)
-                raise MoulinetteError('unable_retrieve_session', exception=error_message)
-            return decrypted.data
+            to_hash = "{id}:{token}".format(id=session_id, token=session_token)
+            hash_ = hashlib.sha256(to_hash).hexdigest()
+
+            if hash_ != stored_hash:
+                raise MoulinetteError('invalid_token')
```
alexAubin, line 189: Note that this is the critical crypto part where we check the user provided id+token against a hashed id+token stored locally during the initial auth.

moulinette/authenticators/__init__.py
```diff
@@ -87,44 +83,57 @@ def __call__(self, password=None, token=None):
...
+                raise
+            except Exception as e:
+                logger.exception("authentication (name: '%s', vendor: '%s') fails because '%s'",
+                                 self.name, self.vendor, e)
+                raise MoulinetteError('unable_authenticate')
+
+            self.is_authenticated = True
+
+            # Store session for later using the provided (new) token if any
+            if token:
```
decentral1se, line 105: We can loop forever without being able to store the session token and keep being asked to authenticate? Should we also raise an exception here to stop such a loop? Unsure if this can happen but trying to follow along all the same :man_technologist:

moulinette/authenticators/__init__.py
```diff
@@ -135,33 +144,46 @@ def _open_sessionfile(self, session_id, mode='r'):
...
+        hash_ = hashlib.sha256(to_hash).hexdigest()
         with self._open_sessionfile(session_id, 'w') as f:
-            f.write(s)
+            f.write(hash_)
 
-    def _retrieve_session(self, session_id, session_hash):
+    def _authenticate_session(self, session_id, session_token):
         """Retrieve a session and return its associated password"""
         try:
+            # FIXME : shouldn't we also add a check that this session file
```
decentral1se, line 160: Perhaps wait for a bug report :smile_cat:

test/conftest.py
```diff
@@ -85,23 +91,51 @@ def patch_logging(moulinette):
...
+def moulinette(tmp_path_factory):
     import moulinette
 
+    # Can't call the namespace just 'test' because
+    # that would lead to some "import test" not importing the right stuff
+    namespace = "moulitest"
+    tmp_cache = str(tmp_path_factory.mktemp("cache"))
+    tmp_data = str(tmp_path_factory.mktemp("data"))
+    tmp_lib = str(tmp_path_factory.mktemp("lib"))
+    os.environ['MOULINETTE_CACHE_DIR'] = tmp_cache
```
decentral1se, line 103: I would recommend to use the `monkeypatch` fixture for patching the environment. Please see https://docs.pytest.org/en/3.0.1/monkeypatch.html. It means that pytest will take care to clean up any changes to the environment for you. Can stop crazy hard to find env bugs later on for someone else.

test/actionsmap/moulitest.yml
```diff
@@ -0,0 +1,42 @@
...
+            configuration:
+                authenticate: false
+
+        default:
+            api: GET /test-auth/default
+            configuration:
+                authenticate: all
+                authenticator: default
+
+#        only-api:
```
decentral1se, line 34: Still needed?

test/conftest.py
```diff
@@ -85,23 +91,51 @@ def patch_logging(moulinette):
...
+@pytest.fixture(scope='session')
+def moulinette_webapi(moulinette):
+
+    namespace = "moulitest"
+
+    api_thread = Process(target=moulinette.api,
+                         args=([namespace],),
+                         kwargs={"host": "localhost", "port": 12342, "use_websocket": False})
+    api_thread.start()
+    time.sleep(0.5)
```
decentral1se: `sleep` in tests is fine at first but can build up (as others see it is "already being used"). Perhaps a while loop with a counter to break out? Things start faster on some machines, slower on others? Alternatively just leave it (so much to be done now anyway) and wait to hear if it causes issues.

tox.ini
```diff
@@ -15,6 +15,9 @@ deps =
   pytest-env >= 0.6.2, < 1.0
   requests >= 2.22.0, < 3.0
   requests-mock >= 1.6.0, < 2.0
+  toml >= 0.10
```
decentral1se: Please add >= and < bounds as it can help to make the test suite more stable going forward.

moulinette/authenticators/__init__.py
randomstuff: It's not super clear to understand what is supposed to be the token, the hash, etc. :)

alexAubin: Eh yup, I added some big comment (dunno if that's the right place for it) to explain what's happening.

test/src/testauth.py
```diff
@@ -0,0 +1,10 @@
+def testauth_none():
```
randomstuff, line 1: What is this file doing?

alexAubin: It's meant to implement dummy functions corresponding to the actionsmap in test/actionsmap/moulitest.yml used for "unit" / functional tests

moulinette/authenticators/__init__.py
```diff
@@ -135,33 +144,62 @@ def _open_sessionfile(self, session_id, mode='r'):
...
-
-        # Encrypt the password using the session hash
-        s = str(gpg.encrypt(password, None, symmetric=True, passphrase=session_hash))
-        assert len(s), "For some reason GPG can't perform encryption, maybe check /root/.gnupg/gpg.conf or re-run with gpg = gnupg.GPG(verbose=True) ?"
+    def _store_session(self, session_id, session_token):
+        """Store a session to be able to use it later to reauthenticate"""
 
+        # We store a hash of the session_id and the session_token (the token is assumed to be secret)
+        to_hash = "{id}:{token}".format(id=session_id, token=session_token)
+        hash_ = hashlib.sha256(to_hash).hexdigest()
```
randomstuff, line 153: Note: could-possibly use the binary SHA-256 instead of converting to hexadecimal.

moulinette/authenticators/__init__.py
```diff
@@ -135,33 +144,62 @@ def _open_sessionfile(self, session_id, mode='r'):
...
+            #      in case this info gets exposed for some reason (e.g. bad file perms for reasons...)
+            #
+            # When the user comes back, we fetch the session_id and session_token from its cookies. Then we
+            # re-hash the {id}:{token} and compare it to the previously stored hash for this session_id ...
+            # It it matches, then the user is authenticated. Otherwise, the token is invalid.
+            #
+            to_hash = "{id}:{token}".format(id=session_id, token=session_token)
+            hash_ = hashlib.sha256(to_hash).hexdigest()
+
+            if hash_ != stored_hash:
```
randomstuff: Would it be useful to use `mac.compare_digest(hash_, stored_hash)`?

moulinette/authenticators/__init__.py
```diff
@@ -31,6 +32,7 @@ class BaseAuthenticator(object):
 
     def __init__(self, name):
         self._name = name
+        self.is_authenticated = False
```
randomstuff, line 36: What is this supposed to do? It is set to `True` at some point and never switched back to `False`. `is_authenticated` is global (shared for all users)?

randomstuff:
I tried to understand what's happening (after the patch) but it's not super clear. Below are some notes of what I think I understood.


`session.id`:

* session ID
* created/set on the login page (if there is not existing session ID)


`session.tokens`:

* signed dictionnary from profile to secret token
* **per-session** secret used to sign `session.tokens` (this secret is stored in process memory)


Secret token:

* random token
* Can be used to login without password?
* How is it used in practice?


Secret token hash:

* SHA-256 `session_id:profile_token`
* Stored in a per-session file (but not per-profile file?).

moulinette/authenticators/__init__.py
```diff
@@ -135,33 +144,62 @@ def _open_sessionfile(self, session_id, mode='r'):
...
+        # We store a hash of the session_id and the session_token (the token is assumed to be secret)
+        to_hash = "{id}:{token}".format(id=session_id, token=session_token)
+        hash_ = hashlib.sha256(to_hash).hexdigest()
         with self._open_sessionfile(session_id, 'w') as f:
-            f.write(s)
+            f.write(hash_)
 
-    def _retrieve_session(self, session_id, session_hash):
+    def _authenticate_session(self, session_id, session_token):
         """Retrieve a session and return its associated password"""
```
Timost: This docstring is outdated. Maybe replace with:
```suggestion
         """Checks session and token against the stored session token"""
```

alexAubin: Indeed :+1:

Wanna make a Github suggestion so that I can commit it and you can get the credit for it ? :stuck_out_tongue_winking_eye: 

(nvm i edited your comment to create the suggestion)

Timost: Thanks ! I'm in the process of reviewing the changes, commit by commit to try to grasp what is going on. '^^

moulinette/authenticators/__init__.py
Timost: I'm not sure but I think it is "switched" off at instantiation and looking at `moulinette/interfaces/api.py:343` it seems that each time an authentication request is made the authenticator is instantiated and therefore this is properly reset to ` False` before any authentication request is made.

moulinette/authenticators/__init__.py
```diff
@@ -135,33 +144,62 @@ def _open_sessionfile(self, session_id, mode='r'):
...
+        to_hash = "{id}:{token}".format(id=session_id, token=session_token)
+        hash_ = hashlib.sha256(to_hash).hexdigest()
         with self._open_sessionfile(session_id, 'w') as f:
-            f.write(s)
+            f.write(hash_)
 
-    def _retrieve_session(self, session_id, session_hash):
-        """Retrieve a session and return its associated password"""
+    def _authenticate_session(self, session_id, session_token):
+         """Checks session and token against the stored session token"""
```
Timost:
```suggestion
        """Checks session and token against the stored session token"""
```
This extra space breaks the tests and the linting

test/test_auth.py
```diff
@@ -0,0 +1,99 @@
...
+    if profile:
+        data["profile"] = profile
+
+    return requests.post(webapi + "/login",
+                         cookies=cookies,
+                         data=data,
+                         headers=None if csrf else {"X-Requested-With": ""})
+
+
+def test_request_no_auth_needed(monkeypatch, tmp_path, moulinette_webapi):
```
Timost: `monkeypatch` and `tmp_path` not used:
```suggestion
def test_request_no_auth_needed(moulinette_webapi):
```

test/test_auth.py
```diff
@@ -0,0 +1,99 @@
...
+
+    r = requests.get(moulinette_webapi + "/test-auth/none")
+
+    assert r.status_code == 200
+    assert r.text == '"some_data_from_none"'
+
+
+def test_request_with_auth_but_not_logged(monkeypatch, tmp_path, moulinette_webapi):
```
Timost: `monkeypatch` and `tmp_path` not used:
```suggestion
def test_request_with_auth_but_not_logged(moulinette_webapi):
```

test/test_auth.py
```diff
@@ -0,0 +1,99 @@
...
+
+    r = requests.get(moulinette_webapi + "/test-auth/default")
+
+    assert r.status_code == 401
+    assert r.text == "Authentication required"
+
+
+def test_login(monkeypatch, moulinette_webapi):
```
Timost: `monkeypatch` not used:
```suggestion
def test_login(moulinette_webapi):
```

test/src/testauth.py
Timost: To add more details. This file and the whole `src/` directory is copied in the `moulinette` (test.conftest.moulinette) fixture for test purposes. You can see usage in the `test.test_auth.test_request_no_auth_needed` for example.

test/conftest.py
Timost: Yes I agree with @decentral1se, although the `scope=session` will mitigate this. But I think there should be other ways to run the http api in the tests than starting it in another thread. Looking at the docs, I stumbled upon [this example](https://bottlepy.org/docs/dev/recipes.html#functional-testing-bottle-applications), which meet the testing needs.

test/conftest.py
Timost: I made a quick test to use `monkeypatch`. Turns out, `monkeypatch` is a `function` scoped fixture and as such, it cannot be used in conjunction with a `session` fixture. However, the `moulinette_webapi` (`session` scoped) uses the `moulinette` fixture and turning it into a `function` scoped fixture makes it do the `time.sleep(0.5)` on every test that uses it '^^.

I think the issue is the `moulinette_webapi` fixture. It should not start the api in a separate thread and should be using something more standard along the lines of [this example from the bottle doc](https://bottlepy.org/docs/dev/recipes.html#functional-testing-bottle-applications).

moulinette/authenticators/__init__.py
Timost: I aggree with @randomstuff using `hmac`'s `compare_diggest` would be better https://docs.python.org/3.6/library/hmac.html#hmac.compare_digest
```suggestion
            if not hmac.compare_digest(hash_, stored_hash):
```

alexAubin:
Wow ô.O thanks a lot for that epic feedback. I need to find some time to depile and carefully read everything.

Trying to provide a quick feedback about your two questions : 
- clearly I don't find my hackish `moulinette_webapi` test thing so great either, but I have no idea how one can test an API. Any tip is much welcome on that part ...
- it's been a long time, but the migration should be completely transparent to the user ? (At least that's an implicit goal I had...)

moulinette/authenticators/__init__.py
alexAubin: Yes ideed, that's significantly better, thanks for that ! (Will `import hmac` in another commit)

moulinette/authenticators/__init__.py
alexAubin: Cheers, that wasn't so clear for me either xD

moulinette/authenticators/__init__.py
alexAubin: Eeeeh sorr I'm answering like 4 months later but can you elaborate on what loop you're thinking about ?

test/actionsmap/moulitest.yml
alexAubin: Hmmm yea idk, the intention was to have a test with `configuration: authenticate: api` where you need to authenticate to access the route, but only if from the API @.@

test/conftest.py
alexAubin: Many thanks for that pointer, looks like the info I was missing to do this cleanly, will have a deeper look

tox.ini
alexAubin:
```suggestion
  toml >= 0.10, < 0.11
```

test/conftest.py
alexAubin: Got reworked in https://github.com/YunoHost/moulinette/pull/216/commits/53ab4709d167ca70b4288a36706de33b1bb60221

test/conftest.py
alexAubin: Got reworked in https://github.com/YunoHost/moulinette/pull/216/commits/53ab4709d167ca70b4288a36706de33b1bb60221

alexAubin:
Soooo that should be pretty much finished, feel free to provide more feedback if you can folks !

PR #217, opened 2019-08-28
Add tests for text/serialize modules

decentral1se:
Also take a pass on the docstrings.

Linking to https://github.com/YunoHost/issues/issues/1255 and https://github.com/YunoHost/issues/issues/1367.


Psycojoker:
LGTM and not that much needing a discussion, thx <3

PR #222, opened 2019-09-16
[ux] make timeout message more explicit

Psycojoker:
Current message is very bad from an UX point of view, people won't understand it so try to make it more explicit.


Psycojoker:
> Should there be a "@Yunohost/translators"? I don't know who these people are sweat_smile

No, everything is done on transifex, github is way way too complicated for them.

PR #224, opened 2019-09-28
Language reworked

comradekingu:


alexAubin:
Yolomergin because those are small legit changes on the README :+1:

PR #226, opened 2019-11-03
[fix] Return nothing if argument is already set.

### LINKED ISSUE #1477: CLI user create: hide password requirements if specified in the command
I suggest to hide the "You are now about to define a new user password..." line if the password has been specified in the `yunohost user create` command.

My system is slow enough that it looked like it was asking to type the password again after displaying this line.

```
yunohost user create test -f John -l Doe -p ThePassword
E-mail address: test@domain.tld
You are now about to define a new user password. The password should be at least 8 characters—though it is good practice to use longer password (i.e. a passphrase) and/or to a variation of characters (uppercase, lowercase, digits and special characters).
TheP << here I thought it was asking for the password
Creating directory '/home/test'.
Success! User created
```

frju365:
## Problem
Even if  the password is in argument of user create (I think it's the same for other commands).

## Solution
Return nothing if arg_value is already set

## How to test
ynh-dev : 
(./ynh-dev start (start container))

To test if behaviors has change **when we don't provide password arg** : 
sudo yunohost user create frju370 --firstname hello --debug

To test how it behaves when **password arg is provided** : 
sudo yunohost user create frju370 --firstname hello -p 00022522104mplo--debug

## Status
**Tested** (need review)

## Misc

fixes https://github.com/YunoHost/issues/issues/1477

### DIFF
--- a/moulinette/actionsmap.py
+++ b/moulinette/actionsmap.py
@@ -89,6 +89,9 @@ class CommentParameter(_ExtraParameter):
     skipped_iface = ['api']
 
     def __call__(self, message, arg_name, arg_value):
+        if arg_value:
+            return
+        
         return msignals.display(m18n.n(message))
 
     @classmethod
--- a/moulinette/actionsmap.py
+++ b/moulinette/actionsmap.py
@@ -91,7 +91,6 @@ class CommentParameter(_ExtraParameter):
     def __call__(self, message, arg_name, arg_value):
         if arg_value:
             return
-        
         return msignals.display(m18n.n(message))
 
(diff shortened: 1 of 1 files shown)


alexAubin:
Uh wokay ?

It can help if you can provide more details on "how to test" than just "ynh-dev" :confused: 

What command did you run ?

frju365:
Ah yes, sorry. I was sure I wrote it. I change it in desc

alexAubin:
Alrighty that work as expected, thanks for this cool fix ! Sorry for the delay

PR #228, opened 2019-12-20
Minimal change to support python3

kay0u:


Psycojoker:
Does it works? I know the ldap library we uses isn't compatible with python3 and we need to switch version.

kay0u:
I'm not sure (and I haven't tested it), no test has been written for the LDAP part. I'm writing tests to cover as much code (including LDAP) as possible.

.travis.yml
```diff
@@ -2,13 +2,13 @@ language: python
 
 matrix:
   include:
-    - python: 2.7
-      env: TOXENV=py27
-    - python: 2.7
+    - python: 3.5
+      env: TOXENV=py35
```
alexAubin: Shouldnt we keep running the tests in 2.7 (this in the lint one ?) as long as we still release a version in 2.7 ?

test/test_filesystem.py
```diff
@@ -33,9 +33,9 @@ def test_read_file_missing_file():
 def test_read_file_cannot_read_ioerror(test_file, mocker):
     error = "foobar"
 
-    with mocker.patch("__builtin__.open", side_effect=IOError(error)):
-        with pytest.raises(MoulinetteError) as exception:
-            read_file(str(test_file))
+    mocker.patch("builtins.open", side_effect=IOError(error))
```
alexAubin: Apparently if we want to be compatible with both python2 and 3, gotta patch something different depending on the python version number :s

kay0u: Yeah. Do we really want to be compatible with both python2 and 3?

alexAubin: I dunno, I was just thinking that would make a smoother transition but if that's too hard we can live without it

kay0u: I tried to keep the compatibility with python 2, the code is ugly however

alexAubin:
Planning to merge this ~soonish in 4.2 because we need to test it on the battlefield

alexAubin:
Yolooooo

PR #230, opened 2019-12-21
Add more tests

kay0u:
Ready for review (good luck)

```
---------- coverage: platform linux2, python 2.7.15-final-0 ----------

Name                                    Stmts   Miss  Cover

-----------------------------------------------------------

moulinette/__init__.py                     39     20    49%

moulinette/actionsmap.py                  311     27    91%

moulinette/authenticators/__init__.py      76     21    72%

moulinette/authenticators/dummy.py         13      0   100%

moulinette/authenticators/ldap.py         127      5    96%

moulinette/cache.py                        12      0   100%

moulinette/core.py                        229     91    60%

moulinette/globals.py                       3      0   100%

moulinette/interfaces/__init__.py         298     99    67%

moulinette/interfaces/api.py              404    138    66%

moulinette/interfaces/cli.py              229     84    63%

moulinette/utils/__init__.py                0      0   100%

moulinette/utils/filesystem.py            174      5    97%

moulinette/utils/log.py                    76     23    70%

moulinette/utils/network.py                25      0   100%

moulinette/utils/process.py                93      9    90%

moulinette/utils/serialize.py              15      0   100%

moulinette/utils/stream.py                 65     12    82%

moulinette/utils/text.py                   28      0   100%

-----------------------------------------------------------

TOTAL                                    2217    534    76%
```


moulinette/utils/filesystem.py
```diff
@@ -290,7 +288,9 @@ def mkdir(path, mode=0o777, parents=False, uid=None, gid=None, force=False):
 
     # Create directory and set permissions
     try:
+        oldmask = os.umask(000)
```
kay0u, line 291: I'm not sure about that

kay0u:
Don't know why import ldif is a problem

kay0u:
Can I do something to make this PR easy to review? Like split this mess into separate PR?

moulinette/utils/filesystem.py
alexAubin: Eh what is it supposed to do ? o.O

moulinette/interfaces/api.py
```diff
@@ -406,13 +404,11 @@ def logout(self, profile):
...
-            # We check that there's a (signed) session.hash available
-            # for additional security ?
-            # (An attacker could not craft such signed hashed ? (FIXME : need to make sure of this))
-            s_secret = self.secrets[s_id]
-            request.get_cookie("session.tokens", secret=s_secret, default={})[profile]
-        except KeyError:
+        # We check that there's a (signed) session.hash available
+        # for additional security ?
+        # (An attacker could not craft such signed hashed ? (FIXME : need to make sure of this))
+        s_secret = self.secrets[s_id]
```
alexAubin: Here I think the point was also to catch the case where `s_id` ain't in `self.secrets` ... Though I'm not sure if that makes any sense, it's been a long time ...

moulinette/interfaces/api.py
```diff
@@ -657,24 +657,19 @@ def add_action_parser(self, name, tid, api=None, **kwargs):
...
+            tid, _ = self._parsers[kwargs.get("route")]
         except KeyError:
-            logger.error("no argument parser found for route '%s'", route)
+            logger.error("no argument parser found for route '%s'", kwargs.get("route"))
             raise MoulinetteError("error_see_log")
 
+        if self.get_conf(tid, "authenticate"):
+            return self.get_conf(tid, "authenticator")
+        else:
+            return False
```
alexAubin:
```suggestion
        return self.get_conf(tid, "authenticate") or False
```

moulinette/interfaces/__init__.py
```diff
@@ -256,45 +256,44 @@ def _validate_conf(self, configuration, is_global=False):
                 raise MoulinetteError("error_see_log")
 
         # -- 'authenticator'
-        try:
+        if "authenticator" in configuration:
             auth = configuration["authenticator"]
```
alexAubin: Can simplify this if/else by 

```python
auth = configuration.get("authenticator", "default")
```

(Maybe done in next commits, will see)

moulinette/interfaces/cli.py
```diff
@@ -446,6 +446,9 @@ def run(self, args, output_as=None, password=None, timeout=None):
         # Set handler for authentication
         if password:
             msignals.set_handler("authenticate", lambda a: a(password=password))
+        else:
+            if os.isatty(1):
+                msignals.set_handler("authenticate", self._do_authenticate)
```
alexAubin, line 460: Uuuuuh, can you elaborate on what this does ? x_X

moulinette/interfaces/api.py
```diff
@@ -657,24 +657,19 @@ def add_action_parser(self, name, tid, api=None, **kwargs):
...
-                return False
-            else:
-                # TODO: In the future, we could make the authentication
-                # dependent of the route being hit ...
-                # e.g. in the context of friend2friend stuff that could
-                # auth with some custom auth system to access some
-                # data with something like :
-                # return self.get_conf(tid, 'authenticator')
-                return "default"
+            tid, _ = self._parsers[kwargs.get("route")]
```
alexAubin, line 668: I'm confused here about removing `route` from mandatory args .... If it's not provided, then it's just gonna try to fetch `self._parsers[None]` which most likely doesn't exist ... What's the motivation ? o.O

moulinette/interfaces/api.py
```diff
@@ -666,7 +666,16 @@ def auth_required(self, args, **kwargs):
...
-            return self.get_conf(tid, "authenticator")
+            authenticator = self.get_conf(tid, "authenticator")
+
+            # If several authenticator, use the default one
+            if isinstance(authenticator, dict):
+                if 'default' in authenticator:
+                    authenticator = 'default'
+                else:
+                    # TODO which one should we use?
+                    pass
```
alexAubin, line 682: Uuuh yeah, not sure what to do here ... All this abstract stuff is fuzzy to me ... Just leaving a comment to remember to check if we shall do something about it ...

moulinette/interfaces/cli.py
```diff
@@ -379,7 +379,16 @@ def auth_required(self, args, **kwargs):
...
-            return self.get_conf(tid, "authenticator")
+            authenticator = self.get_conf(tid, "authenticator")
+
+            # If several authenticator, use the default one
+            if isinstance(authenticator, dict):
+                if 'default' in authenticator:
+                    authenticator = 'default'
+                else:
+                    # TODO which one should we use?
+                    pass
```
alexAubin, line 390: (Same as for api.py)

moulinette/authenticators/ldap.py
```diff
@@ -91,7 +90,12 @@ def authenticate(self, password=None):
             logger.warning("Error during ldap authentication process: %s", e)
             raise
         else:
-            if who[3:] != self.userdn:
+            if (
+                "cn=external,cn=auth" in self.userdn
+                and who[3:] != "cn=admin,dc=yunohost,dc=org"
+            ):
+                raise MoulinetteError("Not logged in with the expected userdn ?!")
+            elif "cn=external,cn=auth" not in self.userdn and who[3:] != self.userdn:
```
alexAubin: Hmmm dunno what to understand here ... Do you still remember and could add comment ?

Also ideally it'd be nice to avoid hard-coded stuff related to yunohost (though that's probably not the only place)

Dunno if that's easy fixable tho...

test/old_slapdtest/README
```diff
@@ -0,0 +1,2 @@
+I have adapted the code from https://github.com/python-ldap/python-ldap/tree/master/Lib/slapdtest since the version of python-ldap we use does not provide the import slapdtest.
+This part will must be removed once we switch to python3
```
alexAubin: u wot m8

alexAubin (commented):
Tried to have a look ... I absolutely ignored what happens in `test/` and only looked at what happens in the "actual" code

Overall it's okay to me, that looks like a pretty interesting work :+1: Would just like to understand better some parts if possible though I'm also willing to take "meh that's just complicated and I don't remember" as an acceptable explanation ¯\\\_(ツ)\_/¯

Congratz on the coverage numbers !

(As a side note, I don't know at all how that relates to other test-related PR pending)

moulinette/utils/filesystem.py
kay0u: without this "fix", "tweak", whatever... the `mode` (permission) of the folder may not be the same as the one given in argument. (depending on current value of the [`umask`](https://wiki.archlinux.org/index.php/Umask)).

But I'm not 100% sure if we want it... I can remove and adapt the tests accordingly

moulinette/interfaces/cli.py
kay0u: If we have a password the first time, we will set the handler "authenticate" to the lambda `lambda a: a(password=password)`. The second time, if we don't have any password I reset the handle the the default behavior `self._do_authenticate` done too in the `__init__` function

moulinette/interfaces/api.py
kay0u: The [parent method](https://github.com/YunoHost/moulinette/blob/4c6945fe0495cbcaa631f64e8f00a3802c2a6606/moulinette/interfaces/__init__.py#L124) doesn't have such argument. My IDE was crying about that. I can either add it again or try catch the route .

Moreover there is a try catch around the value access :-)

moulinette/interfaces/api.py
kay0u: Done

moulinette/interfaces/__init__.py
kay0u: Yeah, sorry --'

moulinette/authenticators/ldap.py
kay0u: I don't know where to put the data if we don't want to hard-code these values.

moulinette/interfaces/api.py
kay0u: Do you want a more "precise" comment here? ¯\\_(ツ)_/¯

moulinette/utils/filesystem.py
alexAubin: Uuuuh wokay ... Not sure to understand but let's roll with it

kay0u: You can find answers here I guess: https://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-in-python

moulinette/interfaces/cli.py
alexAubin: iiiiiiii ... think i'm starting to understand why you ended up doing this, like initializing the moulinette once and willing to call `run()` multiple time ?

Anyway that's sounds legit compared to what's done in `__init__` indeed so marking as resolved

moulinette/interfaces/api.py
alexAubin: Alrighty then xD

moulinette/interfaces/cli.py
kay0u: Yup, that's it! (Maybe we can't do it in real life, but in tests situation we can run multiple time :-°)

moulinette/interfaces/api.py
alexAubin: Hmpf finally was able to re-read it and understand what we're talking about ... It just look like a huge edge case that's never going to happen until we really decide to implement something about this ...

So whatever, that change sounds fine to me :+1:

alexAubin:
https://www.youtube.com/watch?v=yOMj7WttkOA

PR #231, opened 2019-12-22
Don't update LDAP if there are nothing to do

Josue-T:
# Problem

If we try to update the LDAP database with no chang `slapd` crash, and it's a big problem.

# Solution

Check if there are a real change to do and if not do nothing.

### DIFF
--- a/moulinette/authenticators/ldap.py
+++ b/moulinette/authenticators/ldap.py
@@ -232,6 +232,9 @@ class Authenticator(BaseAuthenticator):
         actual_entry = self.search(base=dn, attrs=None)
         ldif = modlist.modifyModlist(actual_entry[0], attr_dict, ignore_oldexistent=1)
 
+        if ldif == []:
+            return True
+
         try:
             if new_rdn:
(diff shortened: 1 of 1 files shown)


Josue-T:
Well it's quite important to merge this also in the actual testing because some unit tests are broken without this PR (because some empty update has been integrated in the [permission factorisation](https://github.com/YunoHost/yunohost/pull/837)).

kay0u (approved):
LGTM, but my slapd don't crash on an empty request... Idk why!

Josue-T:
> LGTM, but I my slapd don't crash on an empty request... Idk why!

But on the testing or stable. I think this crash was introduced by some overlay enabled in the testing.

kay0u:
Yeah, saw it somewhere but I can't reproduce it.

Josue-T:
Actually some unit test are broken following the permission code factorisation. This PR fix this. As I remember it's on the backup/restore part but I don't remember exactly where.

kay0u:
ok, works for me

PR #234, opened 2020-01-14
staticmethod and format_arg_names fix

kay0u:
* staticmethod was not really static
* use `startswith` in format_arg_names

### DIFF
--- a/moulinette/interfaces/__init__.py
+++ b/moulinette/interfaces/__init__.py
@@ -55,7 +55,7 @@ class BaseActionsMapParser(object):
     # Each parser classes must implement these methods.
 
     @staticmethod
-    def format_arg_names(self, name, full):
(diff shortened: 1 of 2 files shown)


moulinette/interfaces/__init__.py
```diff
@@ -156,7 +154,8 @@ def parse_args(self, args, **kwargs):
 
     # Arguments helpers
 
-    def prepare_action_namespace(self, tid, namespace=None):
+    @staticmethod
```
Psycojoker, line 157: Why have you made this method static? It's because "self" isn't used inside?

kay0u: Uuh yes, is it bad?

kay0u: ?

PR #227, opened 2019-12-14
Fix tests breaking due to pytest mock

Timost:
Release 1.12.0 of pytest-mock now raises an error on context manager
usage.
See https://github.com/pytest-dev/pytest-mock/blob/master/CHANGELOG.rst#1120-2019-11-19

PR #236, opened 2020-02-09
Fix tests and linter not already fixed by #227

autra:
Together with #227, this should make the build green.

The root cause of the last failing tests the fact moulinette is writing its lockfiles in `/var/run`. However, at least on Ubuntu, only root can do that. 

This PR fixes that by allowing to configure the runtime dir of moulinette. I chose to add an env variable for this purpose, as it seems more appropriate than in the yml file.

To make things complete, I also added a commit that fixes the linter.


autra:
I just noticed this PR is more or less contained in #230 :-D My solution to the lockfile problem is a bit different though...

kay0u:
Hi, and thank you for your PR. Since https://github.com/YunoHost/moulinette/pull/230 is merged, I think we can close this one.
Feel free to reopen it if necessary.

PR #225, opened 2019-09-29
First steps for py3 migration

decentral1se:
Things I'm unsure about:
  * How to handle the `debian/changelog` file version bump?
  * How to handle the `setup.py` version bump? It's out of sync with ^^^?

This is just one part of this migration. I'll need to manually test this installed against a live YunoHost and probably need to refactor the LDAP code (as far as I remember, the library bindings have changed quite a bit). Maybe I should work in another branch?

I ran:

```
$ 2to3 -f all -f idioms -wn .
```

And then took a manual pass through the code to clean up the rest.

Refs:
  * https://github.com/YunoHost/issues/issues/1255
  * https://github.com/YunoHost/issues/issues/1367

PR #237, opened 2020-02-16
Make test pass with python3.7

autra:
# Description

This PR makes tox succeed with python3.7 (default version on Debian buster). I don't guarantee it does *all* the necessary changes to python3 though. I guess it depends how much we can trust 2to3 to do this job. However, I think it's probably most of what is left to be done, and is actually a good first start.

It depends on:
- ~~#227 and #236 (closed) to fix tests for python2~~
- #228 (python3 bootstrap)

It's basically the followup of #225 (closed now).

Some remarks:
- about the `list()` wrapping in #225, I removed [these](https://github.com/YunoHost/moulinette/pull/225#discussion_r329852847) and [these](https://github.com/YunoHost/moulinette/pull/225#discussion_r329853214) even though @Psycojoker said they were legit. After checking their use, I think an iterable is enough in both cases (but please double-check :-))
- about the eval input [here](https://github.com/YunoHost/moulinette/pull/225#discussion_r329853361): I reverted the change made by 2to3 because I *think* the original `input` should have actually been a `raw_input` (we never prompt user for python code right?) but **please double-check this assumption** as the new python3 behaviour of `input` is very different. 

@alexAubin pinging you as I don't know anyone else to ping :-)

tests to fix:
- [x] test_process.py
- [x] test_filesystem.py
- [ ] test_ldap.py
- [ ] test_auth.py
- [ ] test_actionsmap.py


autra:
Mmmhh ok, lint and format-check actually disagree with each other...

autra:
>  However, I think it's probably most of what is left to be done, and is actually a good first start.

Now that there are more unit tests, I feel I have been a bit naive here :-) Anyway, I'm iterating, tests after tests :-) I'll split this PR later if needed.

autra:
Apparently in python3, `python-ldap` represents text values as str, but attribute values as bytes (cf https://www.python-ldap.org/en/python-ldap-3.2.0/bytes_mode.html). It's not yet clear to me if and where we should convert.

alexAubin:
I'm trying to move forward with Python 3 and dunno what to do with this ... https://github.com/YunoHost/moulinette/pull/228 is more advanced and that's what I based myself on for https://github.com/YunoHost/yunohost/pull/1116 .. is there something we can "salvage" from this PR like tests ?

autra:
>  is there something we can "salvage" from this PR like tests ?

Yes, there are a few things. This PR is (was) actually based on #228, but continued the work. But now that #228 has evolved on its own and merged, it needs a rebase. Are you interested in this rebase? (If so I'll do it)

andretheolauret:
Is this PR still under development? (Maybe can be closed if not)

alexAubin:
Hmpf yeah sorry I doubt we'll be salvaging anything from this PR in the end...

PR #239, opened 2020-04-01
Fix postinstall

kay0u:


moulinette/authenticators/ldap.py
```diff
@@ -94,11 +94,8 @@ def authenticate(self, password=None):
             logger.warning("Error during ldap authentication process: %s", e)
             raise
         else:
-            # If we are trying to login with SASL, we must be logged in as admin
-            if self.sasldn in self.userdn and who != self.admincn:
-                raise MoulinetteError("Not logged in with the expected userdn ?!")
-            # else if the userdn must be the same as the identity
-            elif self.sasldn not in self.userdn and who != self.userdn:
+            # FIXME: During SASL bind whoami from the test server return the admindn while userdn is returned normally :
+            if not (who == self.admindn or who == self.userdn):
```
alexAubin, line 98:
```suggestion
            if not (who == self.admincn or who == self.userdn):
```

?

kay0u: yeah but I think the name `admindn` is more accurate, I'll rename this variable

PR #242, opened 2020-04-23
The hell with 'pythonic try/except' ... Just use a goddamn if/else for that kind of stuff

### LINKED ISSUE #1549: KeyError: 'label' ... where is it coming from !?
For some reason we have this popping at the end of any failed app script ...

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/moulinette/interfaces/api.py", line 331, in wrapper
    curr_v = params[k]
KeyError: 'label'
```

which apparently ain't so important technically, yet people think that's the reason of why their app install/upgrade/.. failed.

Apparently only happens when from the API ?

alexAubin:
This was my initial attempt to fix https://github.com/YunoHost/issues/issues/1549 (until I saw that it still displayed None .. c.f. the root cause discussed in https://github.com/YunoHost/yunohost/pull/952 )

Anyway, I don't understand why people do that kind of try except ... Let's just explicitly say that if the key is not in the dict, we want to do something, and something else otherwise ...

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -327,11 +327,10 @@ class _ActionsMapPlugin(object):
             # Append other request params
             for k, v in request.params.dict.items():
                 v = _format(v)
-                try:
-                    curr_v = params[k]
-                except KeyError:
+                if k not in params.keys():
                     params[k] = v
                 else:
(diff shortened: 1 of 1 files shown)


kay0u (approved):
:+1:

PR #244, opened 2020-04-30
Less spooky messages when session expired

alexAubin:
c.f. https://github.com/YunoHost/issues/issues/1554

Not sure exactly why / when / how but I sometimes see 

```
 Unable to retrieve the session because '[Errno 2] No such file or directory: '/var/cache/moulinette/session/[someid].asc
```

in webadmin during self-upgrades

This PR makes the code a bit smarter and checks the session file exists... Otherwise it most probably means the session expired.


Psycojoker (approved):
LGTM

alexAubin:
Yolomerging (the failed test looks like a weird tmp issue, the other test looks fine, will see)

PR #245, opened 2020-05-01
Simplify interface initialization + ugly optimization hack to speed up parser building

alexAubin:
Ugh I gotta go to sleep but tl;dr: less is moar o/

### DIFF
--- a/moulinette/interfaces/cli.py
+++ b/moulinette/interfaces/cli.py
@@ -431,7 +431,10 @@ class Interface(BaseInterface):
             msignals.set_handler("authenticate", self._do_authenticate)
             msignals.set_handler("prompt", self._do_prompt)
 
-        self.actionsmap = ActionsMap(ActionsMapParser(top_parser=top_parser), load_only_category=load_only_category)
+        self.actionsmap = ActionsMap(
+            ActionsMapParser(top_parser=top_parser),
+            load_only_category=load_only_category,
+        )
 
     def run(self, args, output_as=None, timeout=None):
         """Run the moulinette
--- a/moulinette/interfaces/cli.py
+++ b/moulinette/interfaces/cli.py
@@ -431,7 +431,7 @@ class Interface(BaseInterface):
 
         self.actionsmap = actionsmap
 
-    def run(self, args, output_as=None, password=None, timeout=None):
+    def run(self, args, output_as=None, timeout=None):
         """Run the moulinette
 
         Process the action corresponding to the given arguments 'args'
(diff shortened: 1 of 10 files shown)


kay0u:
https://github.com/YunoHost/moulinette/pull/246

alexAubin:
Demn thanks for fixing the tests for me again xD

moulinette/interfaces/cli.py
```diff
@@ -490,7 +490,11 @@ def _do_authenticate(self, authenticator):
...
         """
-        # TODO: Allow token authentication?
+        # Hmpf we have no-use case in yunohost anymore where we need to auth
+        # because everything is run as root ...
+        # I guess we could imagine some yunohost-independant use-case where
+        # moulinette is used to create a CLI for non-root user that needs to
+        # auth somehow but hmpf -.-
         help = authenticator.extra.get("help")
         msg = m18n.n(help) if help else m18n.g("password")
         return authenticator(password=self._do_prompt(msg, True, False, color="yellow"))
```
alexAubin, line 500: As noted, not sure what to do here ... I was thinking maybe to keep a mechanism using auth through an env variable .. I think that's similar to something I saw for postgresql ... So like 

```
MOULINETTE_PASSWORD="foobar" yunohost some command
```

(though in the context of yunohost we have no need for this, but if we consider moulinette as an independent piece of software it could be useful for some use case) ... I'm not even sure why I wanted to get rid of this "interactive auth" mechanism ... Maybe to make it independent of os.isatty earlier ... but maybe we don't care at all and the current code is fine ...

kay0u: You clearly have a better view of this project than I do, but we're supposed to handle multiple authenticators:
https://github.com/YunoHost/moulinette/blob/d2cb8027927002e06e4309963d593ad3d853090f/test/actionsmap/moulitest.yml#L9-L22

And the tests I've written show that it works as expected. I'm not sure it would be easy to keep this potential feature if we chose to store the password in one file (maybe with different files? but what about ldap authenticator?).

alexAubin: Eh not sure to understand where I was implying a password stored in a file, I was meaning that instead of asking the password interactively it could be provided through an env variable ... But anyway I think we can just keep the current code and it's fine, sorry, that's probably just my talent about overthinking everything :sweat_smile:

kay0u: Yeah me neither, sorry I'm doing 100 things all together :S

Env var? that's a solution indeed :smile:

PR #257, opened 2020-11-22
Support more complex errors (be able to return additional data in a json structure)

alexAubin:
c.f. https://github.com/YunoHost/yunohost/pull/1077 : we want to be able to return additional metadata about an Exception to the API


alexAubin:
Planning to merge this ~soonish in 4.2

PR #258, opened 2020-11-30
[enh] Allow file type in actionmaps

zamentur:
### The problem
There was no way to upload a file through web admin cause the moulinette API don't support multipart forms.

### The solution
Arg parse (on top of which moulinette is built) supports several type to manage file given in arguments (see https://docs.python.org/2.7/library/argparse.html#type ).

If one of this type is specified, we upload the file inside the parsing steps and replace the value by the new temporary path. After what argparse transform the file into an open python file.

At the end of the request the temporary path is deleted automatically by a crazy code using a global var. It was very complex to deal with parsing and abstraction inside moulinette, and i will be happy if you have some suggestion to improve this part of the code.

### Status
Ready for first review

### How to test
You need the PR on user import inside yunohost-admin and yunohost repo. The PR on yunohost-admin use the new vuejs admin.

### Other info
This PR was entirely financed by the new CNAM fablab. Thanks to them :) .


alexAubin:
/me trying to read that PR be like

what-the-fuck-did-i-just-read-what-the-fuck

PR #263, opened 2020-12-31
[fix] Logging

Psycojoker:
See yunohost's PR

https://github.com/YunoHost/yunohost/pull/1119

### DIFF
--- a/moulinette/__init__.py
+++ b/moulinette/__init__.py
@@ -90,12 +90,12 @@ def api(host="localhost", port=80, routes={}):
     except MoulinetteError as e:
         import logging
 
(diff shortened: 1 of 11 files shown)


alexAubin:
I'm a bit confused because in Yunohost we do have many different loggers basically for each file (for example here https://github.com/YunoHost/yunohost/blob/dev/src/yunohost/service.py#L43 ) and they all happen to be logged consistently o.O ?

Psycojoker:
Ah. This is confusing indeed, maybe I've missed some weird stuff in the madness that is the logging configuration '^'

PR #264, opened 2020-12-31
[mod] trash this UX madness that is error_see_logs and improve error messages

Psycojoker:
This "there is an error please look at the logs" error message is an UX abomination and must be removed from earth.

This also try to improve slightly the error messages by integrating at least the exception when it makes sens.


alexAubin:
LGTM :+1:

PR #268, opened 2021-01-25
Fixing tests

kay0u:


moulinette/utils/process.py
```diff
@@ -59,7 +59,7 @@ def call_async_output(args, callback, **kwargs):
         Exit status of the command
 
     """
-    for a in ["stdout", "stderr"]:
+    for a in ["stdout", "stderr", "stdinfo"]:
         if a in kwargs:
             raise ValueError("%s argument not allowed, " "it will be overridden." % a)
```
alexAubin: It's not a huge deal but the reason for this line initially is that stdout or stderr are args later fed to `Popen` (c.f. the `kwargs['stdout'] = ` right after) but stdinfo is not - it's just a custom name that I chose and we don't feed it as a keyword argument to Popen)

alexAubin (approved):
LGTM apart from minor comment

moulinette/utils/process.py
kay0u: Yeah, I was trying more to fix the tests than to understand what going on :+1:

alexAubin:
Can merge when you want :+1: Thanks for all the work !

PR #269, opened 2021-01-26
Fix argparse

kay0u:
2 bug-fixes:

1. `actions_choices` and `subcategories_choices` are OrderedDict, not simple array.
     To reproduce the issue, type `yunohost firewall --help`
     I'll see empty subcategories, but we just want to hide it if it's empty:
     > [...]
     > subcategories:
     > {}
2. There is a bug in argparse.py in python 3.7
     I explained it in comment:
```
# Monkeypatch _get_action_name function because there is an annoying bug
# Explained here: https://bugs.python.org/issue29298
# Fixed by: https://github.com/python/cpython/pull/3680
# To reproduce the bug, just launch a command line without action
# For example:
#       yunohost firewall
# It should display:
#       usage: yunohost firewall {list,reload,allow,disallow,upnp,stop} ... [-h]
#       yunohost firewall: error: the following arguments are required: {list,reload,allow,disallow,upnp,stop}
# But it display instead:
#       Error: unable to parse arguments 'firewall' because: sequence item 0: expected str instance, NoneType found
```


alexAubin (approved):
Eh wokay I suppose ?

(Do you know if that's fixed in Python 3.8+ (asking because recently run some preliminary test on bullseye and was running 3.9 ;P)

kay0u:
> (Do you know if that's fixed in Python 3.8+ (asking because recently run some preliminary test on bullseye and was running 3.9 ;P)

I have no idea :-°

Edit: no it's not on my Manjaro, python 3.9.1

Psycojoker (approved):

PR #276, opened 2021-05-17
[fix] Unicode password doesn't log in

zamentur:
## The problem
Yunohost instance with admin password with an accent or unicde char can't login webadmin / api

https://github.com/YunoHost/issues/issues/1801

## The solution
Bottle is quite strange, if you do request.POST['password'] it differs from request.POST.password ><

https://stackoverflow.com/questions/27432211/python-bottle-requests-and-unicode

## Status
Tested on an lxd ynh-dev container

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -249,7 +249,7 @@ class _ActionsMapPlugin(object):
             def wrapper():
                 kwargs = {}
                 try:
-                    kwargs["password"] = request.POST["password"]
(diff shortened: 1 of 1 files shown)


MCMic:
I’m trying to apply this by hand to fix my installation, what should I restart for the change to take effect ?

zamentur:
```
systemctl restart yunohost-api
```

MCMic:
Thanks a lot, it works!

PR #290, opened 2021-08-30
[enh] Add prefill and multiline in prompt

zamentur:
## The problem
For config panel and the vpn client refactoring, i need to create a question to edit the ovpn template. Here is a solution for the CLI way.

In more, i needed to expose the value already registered in question, so i add the prefill argument instead of a simple '(default: VALUE)' display.

## The solution
Run the default edirot to edit multiline text
Prefill prompt question with the default or current value.

### DIFF
--- a/moulinette/interfaces/cli.py
+++ b/moulinette/interfaces/cli.py
@@ -6,8 +6,11 @@ import getpass
 import locale
 import logging
 import argparse
+import tempfile
+from readline import insert_text, set_startup_hook
 from collections import OrderedDict
 from datetime import date, datetime
+from subprocess import call
 
 from moulinette import m18n, Moulinette
 from moulinette.actionsmap import ActionsMap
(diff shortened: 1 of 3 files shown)


moulinette/interfaces/cli.py
```diff
@@ -533,11 +536,33 @@ def prompt(self, message, is_password=False, confirm=False, color="blue"):
...
+        def prompt(message):
+            if is_password:
+                return getpass.getpass(colorize(m18n.g("colon", message), color))
+            elif is_multiline:
+                while True:
+                    value = input(colorize(m18n.g("edit_text_question", message), color))
+                    if value in ["", "n", "N", "no", "NO"]:
+                        return prefill
+                    elif value in ['y', 'yes', 'Y', 'YES']:
+                        break
```
alexAubin:
```suggestion
                    value = value.lower().strip()
                    if value in ["", "n", "no"]:
                        return prefill
                    elif value in ['y', 'yes']:
                        break
```

PR #312, opened 2021-12-24
Pyupgrade

Psycojoker:
Same than https://github.com/YunoHost/yunohost/pull/1395

### DIFF
--- a/test/test_actionsmap.py
+++ b/test/test_actionsmap.py
@@ -266,7 +266,7 @@ def test_actions_map_import_error(mocker):
     with pytest.raises(MoulinetteError) as exception:
         amap.process({}, timeout=30, route=("GET", "/test-auth/none"))
 
-    expected_msg = "unable to load function {: }.{} because: {}".format(
+    expected_msg = "unable to load function {}.{} because: {}".format(
         "moulitest",
         "testauth_none",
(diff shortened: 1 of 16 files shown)


test/test_actionsmap.py
```diff
@@ -266,7 +266,7 @@ def import_mock(name, globals={}, locals={}, fromlist=[], level=-1):
     with pytest.raises(MoulinetteError) as exception:
         amap.process({}, timeout=30, route=("GET", "/test-auth/none"))
 
-    expected_msg = "unable to load function % s.%s because: %s" % (
+    expected_msg = "unable to load function {: }.{} because: {}".format(
```
alexAubin: Not sure to understand why but that one is making tests crash :thinking: 

```
           expected_msg = "unable to load function {: }.{} because: {}".format(
              "moulitest",
              "testauth_none",
  >           "Yoloswag",
          )
  E       ValueError: Sign not allowed in string format specifier
```

Psycojoker: Ah, maybe the ":" in "{: }"? But that looks weird :/

kay0u: Is there any reason to have a space between `%` and `s`? We could just replace `{: }` by `{}`

PR #324, opened 2022-08-06
[fix] CORS Prefech OPTIONS

zamentur:
## The problem
https://github.com/YunoHost/issues/issues/1533

We had the same issues to integrate swagger API into the documentation of YunoHost

## The solution
It's a partial solutions, cause the hook doesn't work to return cors on HTTP Errors... 
To coturn that we use nginx instead to add the good headers


alexAubin:
Superseded by https://github.com/YunoHost/moulinette/commit/328107c94699074cd00692451fd55589c96f16f2

PR #336, opened 2023-07-14
Fix boring login API expecting a weird form/multiparam thing instead of classic JSON for credentials ...

alexAubin:
Found while working on the next portal API, this just makes stuff unnecessarily painful


alexAubin:
Superseded by https://github.com/YunoHost/moulinette/pull/341

PR #338, opened 2023-08-14
Bypass CSRF protection for the /yunohost/portalapi/login route

selfhoster1312:
Allowing login from simple HTML form... Demo using my_webapp form:

```
<!DOCTYPE html>
<html>
  <head>
    <title>Yunohost SSO</title>
  </head>
  <body>
<?php
    if (array_key_exists("REMOTE_USER", $_SERVER)) {
        echo "Welcome, " . $_SERVER["REMOTE_USER"] . "!";
        echo "<br><a href='/yunohost/portalapi/logout'>Log out<br>";
    } else {
?>
    <form method="POST" action="/yunohost/portalapi/login">
        <input type="text" name="username" id="username">
        <br><input type="password" name="password" id="password">
        <br><input type="submit">
    </form>
<?php
    }
?>
  </body>
</html>
```

Note login and logout route do not redirect anywhere right now so it's not convenient ® yet.

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -272,7 +272,7 @@ class _ActionsMapPlugin:
             name="login",
             method="POST",
             callback=self.login,
-            skip=["actionsmap"],
+            skip=[filter_csrf, "actionsmap"],
         )
         app.route(
             "/logout",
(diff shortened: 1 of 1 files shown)


selfhoster1312:
Note the skipped plugin in the login route is the typed plugin, not the stringy name. Not sure why "actionsmap" is placed as a string there, but when i tried "filter_csrf" (with the quotation marks) it failed miserably.

moulinette/interfaces/api.py
```diff
@@ -362,9 +362,12 @@ def login(self):
             credentials = request.json["credentials"]
             profile = request.json.get("profile", self.actionsmap.default_authentication)
         else:
-            if "credentials" not in request.params:
+            if "credentials" in request.params:
+                credentials = request.params["credentials"]
+            else:
+                if "username" in request.params and "password" in request.params:
+                    credentials = request.params["username"] + ":" + request.params["password"]
```
alexAubin, line 369:
```suggestion
            if "credentials" in request.params:
                credentials = request.params["credentials"]
            elif "username" in request.params and "password" in request.params:
                credentials = request.params["username"] + ":" + request.params["password"]
```

selfhoster1312:
I can't change my branch in the PR so please forget this one and look at #340 instead

PR #340, opened 2023-08-14
portal-api: Bypass CSRF protection for login route

selfhoster1312:
Allowing login from simple HTML form
Also allow to pass username/password as two params instead of a combined "credentials"

Demo using my_webapp:

```
<!DOCTYPE html>
<html>
  <head>
    <title>Yunohost SSO</title>
  </head>
  <body>
<?php
    if (array_key_exists("REMOTE_USER", $_SERVER) && $_SERVER["REMOTE_USER"] != "") {
        echo "Welcome, " . $_SERVER["REMOTE_USER"] . "!";
        echo "<br><a href='/yunohost/portalapi/logout?referer_redirect'>Log out<br>";
    } else {
?>
    <form method="POST" action="/yunohost/portalapi/login?referer_redirect">
        <input type="text" name="username" id="username">
        <br><input type="password" name="password" id="password">
        <br><input type="submit">
    </form>
<?php
    }
?>
  </body>
</html>
```

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -272,13 +272,14 @@ class _ActionsMapPlugin:
             name="login",
             method="POST",
             callback=self.login,
-            skip=["actionsmap"],
+            skip=[filter_csrf, "actionsmap"],
         )
         app.route(
             "/logout",
             name="logout",
             method="GET",
             callback=self.logout,
+            # No need to bypass CSRF here because filter allows GET requests
             skip=["actionsmap"],
         )
 
@@ -362,9 +363,12 @@ class _ActionsMapPlugin:
             credentials = request.json["credentials"]
             profile = request.json.get("profile", self.actionsmap.default_authentication)
         else:
-            if "credentials" not in request.params:
+            if "credentials" in request.params:
+                credentials = request.params["credentials"]
+            elif "username" in request.params and "password" in request.params:
+                credentials = request.params["username"] + ":" + request.params["password"]
+            else:
                 raise HTTPResponse("Missing credentials parameter", 400)
-            credentials = request.params["credentials"]
 
             profile = request.params.get("profile", self.actionsmap.default_authentication)
 


moulinette/interfaces/api.py
```diff
@@ -272,13 +272,14 @@ def setup(self, app):
             name="login",
             method="POST",
             callback=self.login,
-            skip=["actionsmap"],
+            skip=[filter_csrf, "actionsmap"],
```
alexAubin, line 275: I'm kinda confused ... I understand that basic, "raw" HTML form won't include the special X-Requested-With header ... but it looks like disabling this checks opens up the very security issue that this CSRF is supposed to prevent ... or would a CSRF token help fix the "raw" HTML form support, rather than relying only on X-Requested-With ?

selfhoster1312: My understanding: CSRF is a problem where you steal credentials from an authenticated user from another webpage. On the login/logout route, it should not have any security implications.

selfhoster1312:
About CSRF in general, it's desirable to protect from such attacks. However, a header token is not the only approach.

For actions where the user should be logged in, we can:

- generate a unique token server side to be provided in the HTML form, what's apparently called [synchroniser token pattern](https://en.wikipedia.org/wiki/Cross-site_request_forgery#Synchronizer_token_pattern) ; not sure how effective that is, can't the attacker just read the token by making a request for the form first?
- set a stricter cookie policy ; currently Yunohost uses [samesite=lax](https://github.com/search?q=org%3AYunoHost+SameSite&type=code) so that the user is not authenticated at all in case of CSRF

About the cookie policy, is it lax only because SSOWat needs to handle different domains? If so maybe we could workaround by having a cookie per SSOWat domain?

moulinette/interfaces/api.py
alexAubin: Alrighty:

>Definition. Cross-Site Request Forgery (CSRF) is an attack that forces **authenticated** users to submit a request to a Web application **against which they are currently authenticated**.

PR #342, opened 2023-09-27
Bookworm + portal rework

alexAubin:


selfhoster1312:
I don't think a6c7e55d1dcfc26c94cc97713b270037d236b30a fixes https://github.com/YunoHost/issues/issues/2238

The cookie should be deleted by the client but other copies of it (or if the client doesn't honor the request) remain valid for SSOWat... or did i miss something? So that's more like "delete cookies" than "logout" effectively, which i think is a security problem.

EDIT: This should have been fixed in the meantime

selfhoster1312:
I think all the building blocks are here to have a portal without javascript. I'll try and let you know. Thanks a lot!!!

selfhoster1312:
Soooo after messing around a little bit i realized we don't have the building blocks at all: yunohost-portal-api has no privileged API for a backend/daemon to know what permissions a user has, which was possible previously with `/etc/ssowat/ssowat.json` (although it was a hack and not intended). I opened #2433 to discuss it

PR #347, opened 2024-03-07
ci(autoblacks): add step to check if black reformatted some code

Psycojoker:
Because psf black job is a bit dumdum.


Psycojoker:
It works (and fix the broken workflow), yolo.

PR #359, opened 2024-07-12
API - Add support for request content in JSON

Josue-T:
# Problem

Using yunhost admin API with as REST API has not issue.

By example when we use it with the requests lib passing the parameter is straightforward. And I didn't find a working solution to pass correctly an `UTF-8` easily with `requests`.

# Solution

Add a working support for REST request with JSON content type.

Note that for now the authentication is not supported by this way. But the issue currently is on the other request which could be difficult to write to make it working correctly.

# How to test

By example we requests we can test this way:

```python3
import requests
s = requests.Session()
s.headers['X-Requested-With'] = 'Client'
s.post("https://domain.tld/yunohost/api/login", data={"credentials": "%s:%s" % ("user", "pwd")})
s.post("https://domain.tld/yunohost/api/users", json={"username": "yolotest", "fullname": "Fullname Withspécialchàr", "password": "pwd", "domain": "domain.tld"})
```

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-
+import json
 import re
 import errno
 import logging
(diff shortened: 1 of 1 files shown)


alexAubin:
Closing because supposedly it's now live in bookworm : https://github.com/YunoHost/moulinette/blob/dev/moulinette/interfaces/api.py#L313

PR #373, opened 2025-05-11
Support nargs in URL parameters in Ynh API

christophehenry:
This add support for [nargs argument in route parameters](https://github.com/YunoHost/yunohost/blob/dev/share/actionsmap.yml#L647-L657). They should be separated by space, URL-encoded (i.e. separated by `%20`)

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -19,6 +19,7 @@
 #
 
 import os
+import shlex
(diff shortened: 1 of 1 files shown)


christophehenry:
Deleting this one as not relevant. This was more of an experiement than a real proposition. Feel free to reopen if someone actually needs it.

PR #380, opened 2026-06-25
[fix] String started with at was interpreted as option

zamentur:
## The problem
Text related to an API option and starting with an at are interpreted as options and not as values !

Related issues: https://github.com/YunoHost/issues/issues/2557

## The solution
It's an argparse bug https://github.com/python/cpython/issues/138950

Some comments suggest to use the equal notation but it doesn't work for positional arguments...

SO i found a dirty workaround cause no easy way to escape the value exists (doesn't really make sens in the usual context of argparse i guess...). I add a space at the beggining of the string and remove it just at the end of the wrapper function.

**AI transparency:** no AI used
## PR Status

Tested manually

## Code TODOs
seems ready

## Tests TODOs

Indicate here tests you have already done, and tests you think should be done

 - [x] Call the code through webadmin call by hand and use a logger.warning to get the value of the return
 - [x] Change a user password starting with @ and log with it
 - [ ] Test to upload a file
 - [x] Test an option allowing several values as a list

### DIFF
--- a/moulinette/interfaces/api.py
+++ b/moulinette/interfaces/api.py
@@ -231,16 +231,14 @@ class _HTTPArgumentParser:
                 value.save(UPLOAD_DIR)
                 if option_string is not None:
                     arg_strings.append(option_string)
+                arg_strings.append(f"{UPLOAD_DIR}/{value.filename}")
+            elif isinstance(value, str):
                 # ==== SPACE DIRTY_HACK =====
                 # In order to avoid option injection by starting a value with a
-                # prefix_chars (here @), we add a starting space on all string
-                # values, like that argparse do not interpret it as an option...
-                # The space is removed in a post-treatment
-                # see https://github.com/python/cpython/issues/138950
-                arg_strings.append(f" {UPLOAD_DIR}/{value.filename}")
-            elif isinstance(value, str):
-                # Same space dirty hack as above
-                protected_value = f" {value}"
+                # prefix_chars (here @), we add a starting space on string values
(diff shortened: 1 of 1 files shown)


moulinette/interfaces/api.py
```diff
@@ -227,21 +227,33 @@ def append(arg_strings, value, action):
                 value.save(UPLOAD_DIR)
                 if option_string is not None:
                     arg_strings.append(option_string)
-                arg_strings.append(UPLOAD_DIR + "/" + value.filename)
+                # ==== SPACE DIRTY_HACK =====
+                # In order to avoid option injection by starting a value with @
+                # we add a starting space on all string values, like that argparse
+                # do not interpret it as an option...
+                # The space is remove in a post-treatment
```
tituspijean:
```suggestion
                # The space is removed in a post-treatment
```

moulinette/interfaces/api.py
```diff
@@ -269,7 +281,13 @@ def append(arg_strings, value, action):
             if dest in args:
                 arg_strings = append(arg_strings, args[dest], action)
 
-        return self._parser.parse_args(arg_strings, namespace)
+        # Post-treatment of the dirty hack describe above
+        parsed_args = self._parser.parse_args(arg_strings, namespace)
+        for option_string, value in vars(parsed_args).items():
+            if isinstance(value, str) and option_string not in ["loginShell", "_tid"]:
```
alexAubin: Wtf why "loginShell" and "_tid" ? This needs a comment x_x

